Issue #345, opened 2026-07-23
codegen --format wit emits opaque list<u8> for every AADL data type — how to get a structured WIT record/list<f32> for an inter-core message?


avrabe:
Using `spar codegen --format wit` to derive the inter-core WIT contracts from an AADL model (jess Pixhawk 6X-RT). Every AADL `data` type maps to `type X = list<u8>` regardless of structure — I can't find the construct that produces a **typed** WIT (a `record`, or `list<f32>`). The v0.30 changelog mentions a data-plane feature (REQ-CODEGEN-WIT-DATAPLANE-001, "generated Guest marshals typed data across the WIT ABI"), so this may be a "how do I declare it" question rather than a gap — but two natural AADL constructs both yield opaque bytes.

## Repro (spar v0.24.0 AND v0.30.0, identical result)
A port `state_out : out data port VehicleState` where VehicleState is enriched either way:

**(a) Data_Model properties:**
```aadl
data VehicleState
  properties
    Data_Model::Data_Representation => Array;
    Data_Model::Base_Type => (classifier(Base_Types::Float));
    Data_Model::Dimension => (16);
end VehicleState;
```
**(b) data implementation with subcomponents:**
```aadl
data VehicleState end VehicleState;
data implementation VehicleState.nav
subcomponents
  qw : data Base_Types::Float;  qx : data Base_Types::Float;
  qy : data Base_Types::Float;  qz : data Base_Types::Float;
end VehicleState.nav;
```
Both produce, in the generated WIT:
```wit
type vehicle-state = list<u8>;
set-state-out: func(val: vehicle-state);
```
(Opaque `list<u8>` — the Float/Array/Dimension and the subcomponent fields are dropped.)

## jess impact
For REQ-PIX-007 (inter-core WIT-typed contracts on the RT1176), the M4→M7 crossing is a **NavState = 16×f32 (64 B)** message. We've proven it crosses byte-exact on the dual-core Renode model, but the *contract* can only say `list<u8>`, not `record navstate { qw: f32, ... }` or `list<f32>` — so the typed ABI can't reflect the real payload, and consumers can't be type-checked against it.

## Suggested mapping (if not already supported)
- `data implementation` **subcomponents** → WIT `record` (one field per subcomponent; `Base_Types::Float`→`f32`, `Integer`→`s32`, `Unsigned_32`→`u32`, nested data→nested record).
[...]

Issue #327, opened 2026-07-11
Certify AADL↔WIT↔canonical-ABI byte-layout equivalence with ordeal (close the #319 layout gap)


avrabe:
spar's codegen maps an AADL `data implementation` → WIT `record` → wasm canonical-ABI byte layout, but the width/offset table (`spar-codegen/src/wit_gen.rs`, `spar-transform/src/wit.rs`) is **syntactic and never checked** — #319 shows a 9-byte `EepromSnapshot` collapsing to an opaque `list<u8>`.

**Proposal:** use **ordeal** `prove_equiv` to prove each generated encoding faithfully represents the source layout (field offsets/widths, enum/tag discriminants, scalar padding/alignment) — the "does a WIT record faithfully encode an AADL data impl" check, currently unimplemented.

Enablers: byte-layout helpers — ordeal roadmap **pulseengine/ordeal#64** (v0.10.0) — + the equivalence toolkit **pulseengine/ordeal#66** (v0.12.0). Note: spar has no ordeal usage today (this is greenfield; #38 is the Lean scheduling kernel, not layout). **Boundary:** byte-layout is QF_BV; deployment MILP, RTA/EDF, TSN/WCTT, ARINC 653 are LP/real arithmetic and stay on HiGHS/good_lp.

avrabe:
**Triage (first pass).** Confirmed real gap, not closeable. The AADL `data impl` → WIT `record` → canonical-ABI width/offset table (`spar-codegen/src/wit_gen.rs`, `spar-transform/src/wit.rs`) is syntactic and unchecked; #319's 9-byte `EepromSnapshot` collapsing to `list<u8>` is the concrete symptom.

**Classification:** actionable correctness-certification, but **externally blocked**. It depends on ordeal byte-layout helpers (pulseengine/ordeal#64, v0.10.0) + the equivalence toolkit (pulseengine/ordeal#66, v0.12.0), and spar has zero ordeal usage today (greenfield integration). Boundary is right: byte-layout is QF_BV for ordeal; the LP/real solvers (RTA/EDF/TSN/ARINC, HiGHS/good_lp) stay put.

[...]

PR #346, opened 2026-07-23
feat(codegen): emit typed WIT list from Array data types (#345)

avrabe:
## Summary

Closes the #345 gap: `spar codegen --format wit` mapped **every** AADL `data` type with `Data_Representation => Array` to an opaque `type x = list<u8>`, dropping the element type and dimension — so an inter-core WIT contract could only say `list`, not the real payload (the jess NavState `16 × f32`).

This teaches codegen to emit a **typed** WIT list from a Data Modeling Annex `Array` data type.

## What changed

- **`spar-hir-def` (`resolver.rs`)** — new `DataShape::Array { element, len }`. `resolve_data_shape` recognizes `Data_Representation => Array`, resolves the `Base_Type` classifier to the element shape (recursively, reusing the existing scalar/record resolution), and reads the first `Dimension` as the length. Both the maintainer's list-wrapped property syntax (`Base_Type => (classifier(Pkg::T))`, `Dimension => (16)`) and the bare forms are accepted.
- **`spar-codegen` (`wit_gen.rs`)** — a fixed `Dimension` emits a **bounded** fixed-size `list<T, N>` (no `cabi_realloc`, consistent with the RECORDS‑002 no_alloc discipline where only unbounded `list`/`string` force dynamic allocation); no `Dimension` emits `list<T>`. The element resolves through a shared `shape_wit_type` helper, so an array-of-record and an array-typed record **field** both work.
- **`spar-codegen` (`rust_gen.rs`)** — the Rust port type mirrors wit-bindgen's lowering: `list<T, N>` → `[T; N]`, `list<T>` → `Vec<T>`.

## Honest scoping / no regression

An `Array` whose `Base_Type` is absent **or** resolves to an unsized/unresolvable element (e.g. the standard `Base_Types::Float`, which carries no `Data_Size`) is **not fabricated**: `array_of` returns `None`, so a top-level array port **gracefully keeps the legacy `list<u8>`** (no regression, no hard error), and an array-typed record field takes the record no_alloc hard-error path — exactly as an opaque field would.

Not claimed here: `list` byte-layout ↔ canonical-ABI equivalence (that's #327's ordeal work); multi-dimensional arrays use only the first `Dimension`; no compiled-crate oracle for array *ports* on the Rust side.

## Oracles (executed unit tests, `spar-codegen/src/wit_gen.rs`)

- #345 NavState → `type vehicle-state = list<f32, 16>;`
- unbounded array → `type samples = list<f32>;`
- no `Base_Type` → `type mystery = list<u8>;` (negative control)
- unsized element → `type blurry = list<u8>;` (graceful fallback, no error)
- array-typed record field → `nav: list<f32, 16>`

Each asserts the emitted WIT **parses + resolves** under `wit_parser` (a real fixed-size-list bind, not a string match).

`cargo test -p spar-codegen --lib` → 56 passed. `cargo test -p spar-hir-def --lib` → 471 passed. `cargo fmt --check` clean; `cargo clippy` on both crates clean. An independent clean-room agent re-derived each claim against its own fixtures.

Tracked as **REQ-CODEGEN-WIT-ARRAY-001**.

Note: case (b) of #345 (a `data implementation` with subcomponents) already works today when the port references the **implementation** classifier (`VehicleState.nav`) rather than the bare `data` type — details in the issue reply.

---
_Generated by [Claude Code](https://claude.ai/code/session_016yT1pAR89FsY6zRpC48Bzj)_

### DIFF
--- a/artifacts/requirements.yaml
+++ b/artifacts/requirements.yaml
@@ -3055,6 +3055,58 @@ artifacts:
       - type: traces-to
         target: REQ-CODEGEN-WIT-STATE-001
 
+  - id: REQ-CODEGEN-WIT-ARRAY-001
+    type: requirement
+    title: "Structured WIT list from a Data Modeling Annex Array data type"
+    description: >
+      Codegen WIT emission shall map an AADL `data` type whose Data Modeling Annex
+      `Data_Representation` is `Array` to a typed WIT `list`, closing the #345 gap
+      where EVERY array data type degraded to the opaque `type x = list<u8>` blob
+      (element type, signedness/float, and dimension all dropped), so an inter-core
+      contract could only say `list`, not the real payload. Delivered and
+      oracle-verified:
+      (1) `resolve_data_shape` gains a `DataShape::Array { element, len }` arm
+      (spar-hir-def): a `data` type with `Data_Representation => Array` resolves its
+      `Base_Type` classifier to the element shape (recursively, reusing the same
+      scalar/record resolution as records) and reads the first `Dimension` value as
+      the length. Both the maintainer's list-wrapped property syntax
+      (`Base_Type => (classifier(Pkg::T))`, `Dimension => (16)`) and the bare forms
+      are accepted (typed_value: `List([ClassifierValue])`/`ClassifierValue`,
+      `List([Integer])`/`Integer`).
+      (2) A FIXED `Dimension` emits a BOUNDED fixed-size WIT `list<T, N>` — no
+      `cabi_realloc`, consistent with the RECORDS-002 no_alloc discipline (only
+      unbounded `list`/`string` force dynamic allocation); an array with NO
+      `Dimension` emits an unbounded `list<T>`; an `Array` with NO resolvable
+      `Base_Type` is NOT fabricated — it keeps the legacy opaque `list<u8>`.
+      (3) The element resolves through the shared `shape_wit_type` helper, so an
+      array-of-record and an array-typed record FIELD both work; the generated Rust
+      port type mirrors wit-bindgen's lowering (`list<T, N>` -> `[T; N]`,
+      `list<T>` -> `Vec<T>`).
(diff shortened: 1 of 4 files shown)


avrabe:
## CI blocked by self-hosted runner disk exhaustion (infra, not this diff)

Three required checks failed — **Proptest (extended)**, **Fuzz smoke**, **Bench compile smoke** — all with the same root cause: the self-hosted runners are **out of disk**.

- `Bench compile smoke` (runner5): `error: failed to write to '.../target/release/deps/rmetaNfJLue/full.rmeta': No space left on device (os error 28)`
- `Proptest (extended)` (runner11): `ld terminated with signal 7 [Bus error], core dumped` while linking `spar-mcp` (SIGBUS = the linker's mmap'd output hit a full disk)
- Both post-job hooks report `disk: 100% used (threshold 70%)`, and the cleanup step could not reclaim space (`cleanup done: disk 100% → 100%`).

This is **not related to this PR's change**: it's isolated to `spar-mcp` linking / `lsp-types` codegen (crates this diff doesn't touch), and `spar-codegen` / `spar-hir-def` (the crates I changed) compiled cleanly earlier in the same logs. Locally: `spar-codegen` 56/56, `spar-hir-def` 471/471, fmt + clippy clean.

Two runners (runner5, runner11) are both at 100% — this looks fleet-wide, which likely also explains why the `main` `v0.32.0` CI run has been stuck `queued` for hours.

**Action needed (runner admin):** reclaim disk on the self-hosted runner fleet (prune `target/`, sccache, and `~/.cargo/registry/cache` — the current post-job hook isn't freeing enough). The required checks can't go green until then. I don't have permission to re-run failed jobs (`rerun-failed-jobs` → 403), and a re-trigger won't help while the runners are full. I'll re-verify and drive this to merge once the disks are back under threshold — nothing to change on the branch itself.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #358, opened 2026-07-30
Verification Gate scope is stale: 20 of 161 feature artifacts checked, per-PR override never used


avrabe:
## Summary

The Verification Gate's scope has been stale since ~v0.11.0. It re-runs the same
20 v0.9.3/v0.10.0-era artifacts on every PR and never executes the artifacts the
PR under review actually adds or changes. The per-PR escape hatch designed to fix
this has never once been used.

Consequence: for **141 of 161** `type: feature` artifacts, the V-model's right
side is not mechanically closed. An artifact can be committed as
`status: passing` / a requirement as `status: verified` without its declared
`fields.steps[].run` ever having been executed by CI.

## Evidence

`.github/workflows/verification-gate.yml:66` — the default filter:

```
DEFAULT='(and (= type "feature") (or (has-tag "v093") (has-tag "v0100")))'
```

Measured against `artifacts/*.yaml` at `c065e3d`:

| | count |
|---|---:|
| `type: feature` artifacts | 161 |
| matched by the default filter | **20** |
| never executed by the gate | **141** |

The workflow intends each PR to narrow the scope to its own artifacts with a
`Verify-Filter:` body line (`verification-gate.yml:52-72`). Checking the last ten
merged PRs — #354, #353, #352, #351, #349, #348, #346, #344, #342, #340 — **none
set it.** Every one ran against the stale 20.

The clearest illustration is #356, which *removed* `TEST-LAYOUT-CERT` and rewrote
three requirements. Its gate comment:

> ✅ **20/20** passed
> **Filter:** `(and (= type "feature") (or (has-tag "v093") (has-tag "v0100")))`

A green check that verified twenty unrelated artifacts and none of the PR's own.

## Why the existing escape hatch cannot work

`Verify-Filter:` is read from `github.event.pull_request.body`, which is captured
in the event payload **at trigger time**. So:

- adding the line after opening the PR has no effect on the run already fired;
- re-running the workflow replays the original payload, so it does not pick up an
  edited body either;
- it only takes effect on the next `synchronize`.

A gate scope that depends on prose typed into a description *before* opening the
PR, and that silently no-ops if forgotten, is not a gate. This is the same shape
as the required-check hole fixed in #353: the failure mode is *silence that looks
like success*.

## Proposed fix

Make the scope a function of the commit instead of the description.

1. Add `tools/changed_verification_artifacts.py`: parse
   `git show <merge-base>:artifacts/verification.yaml` and the working copy, diff
[...]

PR #359, opened 2026-07-30
ci(guard): enforce human-scoped artifacts mechanically (REQ-GUARD-HUMAN-SCOPED-001)

avrabe:
## What

Makes the `human-scoped` designation **mechanical**. An artifact tagged
`human-scoped` may sit at `draft`/`proposed`/`approved` and may be
`rejected`/`deprecated`/`withdrawn`, but may not be recorded
`implemented`/`verified`/`released`/`accepted`/`passing`. To ship one, delete the
tag **in the same change** — the promotion becomes one explicit line of diff
instead of a status buried in a large PR.

New: `REQ-GUARD-HUMAN-SCOPED-001` (`release: v0.36.0`), `TEST-GUARD-HUMAN-SCOPED`,
`tools/check_human_scoped.py`, three fixtures.

## Why

v0.35.0's `REQ-CODEGEN-LAYOUT-CERT-001` was designated human-scoped **in a
planning note only**. The autonomous issue-hunt cron picked it up, implemented it,
self-recorded `status: verified`, and merged it in about an hour — asserting a
proof that did not exist, because the SMT goal constant-folded and every input
therefore produced a byte-identical certificate. Nothing anywhere enforced the
designation. A guardrail that lives only in prose is not a guardrail.

### Counterfactual, run against the live #354 state

`REQ-CODEGEN-LAYOUT-CERT-001` is *still* `status: verified` on `main` until #356
lands, so this is a real run, not a reconstruction:

```
error: 1 'human-scoped' artifact(s) record a completed status.
  REQ-CODEGEN-LAYOUT-CERT-001: status=verified  (requirements.yaml)
EXIT=1
```

It would have turned the **required** `rivet-validate` context red.

## Honest ceiling — read before trusting this

**A speed bump, not an authorization boundary.** `main` requires zero approving
reviews, and that is *forced* rather than sloppy: with `enforce_admins: true` on a
single-maintainer repo, GitHub's ban on self-approval makes any required review an
unbreakable deadlock. Commit signatures can't discriminate either, since
squash-merges are re-signed by GitHub's own key (`key=B5690EEEBB952194`, web-flow).

So an agent with push access can delete the tag in the same PR and green this
check. What the check buys is that doing so becomes **explicit, diffable and
logged** rather than invisible. The only real gate is on the agent's prompt and
permissions, which lives outside this repo. This is stated in the script's own
docstring, in the requirement, and in AGENTS.md, so it can't quietly be read as
stronger than it is.

### Three exit codes, not two

`0` clean · `1` violated · `2` **INCONCLUSIVE**.

Conflating "the property is violated" with "the check could not run" is the same
category error as the bus-bandwidth defect in #357, where `0.0` meant both *no
bandwidth* and *unmeasurable*. Same fix shape, one layer up.

### Stdlib-only, and fail-closed

No workflow in this repo installs a Python package, and every other `tools/*.py`
is stdlib-only (`python3` itself is proven present — `verification-gate.yml:80`
invokes `run_verification.py` by shebang). Importing PyYAML would make a
**required** check depend on an unverified runner package, and *a guardrail that
cannot run reads as approval*.

So it parses the narrow subset the artifacts actually use, and **refuses anything
else**: block-form `tags:` exits 2 rather than being read as "no tags". A
best-effort scanner would degrade to a false green — the exact vacuity that let
#354 through.

Equivalence was **measured, not assumed**: differentially identical to PyYAML
across all **886** committed artifacts (`artifacts/` + `safety/stpa/`) — same id
sets, same `(status, tags)` for every artifact, **zero divergence**.

### Non-vacuity is self-checking

`--self-test` runs three committed fixtures and asserts each exit code, **before**
the real scan, so a toothless refactor fails loudly instead of letting the real
scan pass:

| fixture | must exit | proves |
|---|---|---|
| `human-scoped-violation` | 1 | two artifacts differing **only** in status → discriminates on status, not the tag |
| `human-scoped-unsupported` | 2 | block-form `tags:` refused, not read as "no tags" (which would silently *exempt* it) |
| `human-scoped-duplicate-id` | 1 | a duplicate id cannot shadow a violation |

## A real defect found while building this: six duplicate artifact ids

The duplicate-id fixture is not hypothetical. `RENDER-REQ-001`…`006` are each
defined **twice** — in `artifacts/requirements.yaml` *and* in
`safety/stpa/rendering-analysis.yaml`, both in rivet's scope (`rivet.yaml` lists
`artifacts` and the whole `safety/stpa` directory). The copies have **drifted**:

| id | `artifacts/requirements.yaml` | `safety/stpa/rendering-analysis.yaml` |
|---|---|---|
| RENDER-REQ-003 | `implemented` | `partial` |
| RENDER-REQ-006 | `partial` | `implemented` |

Contradicting each other in opposite directions, plus two divergent titles. **rivet
v0.4.3 — the version pinned in the required gate — validates all of that green.**

Hence the checker iterates **records**, never a `{id: record}` dict. Verified by
mutation: dedup-by-id flips that fixture 1→0 and reddens the self-test, while
leaving the other two fixtures unaffected — so the fixture carries unique
detection weight.

The duplicate ids themselves are **pre-existing and filed separately**, not fixed
here: choosing which copy is canonical is a content decision about the rendering
requirements, not part of a guardrail PR.

## `.gitignore` lines 13–15 were unanchored

`requirements.yaml` without a leading `/` matches that basename at **any depth**,
which silently excluded the new fixtures:

```
$ git add tools/fixtures/          # EXIT=0, nothing staged   <- silent
$ git add tools/fixtures/.../requirements.yaml
The following paths are ignored by one of your .gitignore files:   # EXIT=1 <- loud
```

`ee2c831` added them as *"dist/ and loose yaml"* — i.e. root-level scratch — so
`/`-anchoring is a faithful fix, not a policy change. Nothing else on disk matches
either form. The tracked `artifacts/` and `safety/stpa/` copies were unaffected
only because gitignore does not apply to already-tracked files.

## Artifacts tagged in this PR

Three, all at `proposed`, so the check passes while genuinely asserting something:

- **`REQ-BMC-CONCURRENCY-001`** — the v0.1 BMC spike needs a human-designed
  falsification kill-criterion and an honest practical-`k` measurement.
- **`REQ-PROOF-NC-MINPLUS-001`**, **`REQ-PROOF-NC-CERT-001`** — research-grade
  proof obligations needing human narrowing to something a solver can decide.

`REQ-PROOF-RTA-OVERFLOW-001` and the withdrawn `REQ-CODEGEN-LAYOUT-CERT-001` get
the tag once #356 lands (tagging the latter *now* would legitimately redden CI,
since it is still `verified` on `main`).

Note the checker warns `no artifact is tagged 'human-scoped', so this check
asserted nothing` when the set is empty — it refuses to print a reassuring blank.

## Verification

- `tools/check_human_scoped.py --self-test` — **3/3 pass**, cwd-independent.
- Real scan, exactly as CI runs it — **EXIT=0**, 886 artifacts, 3 tagged, all `ok`.
- Differential vs PyYAML over both artifact roots — **IDENTICAL**, 0 divergence.
- Mutation test (dedup-by-id) — self-test correctly goes **red**.
- **rivet delta measured against pristine `main` c065e3d in an isolated
  worktree**, not assumed: errors **465 → 466**, warnings **624 → 624**, lifecycle
  coverage gaps **102 → 103**. The one new error is `TEST-GUARD-HUMAN-SCOPED`'s
  `status: passing` — the repo convention for `type: feature` that CI's pinned
  rivet v0.4.3 accepts and local 0.28.0 rejects, matching **275** existing
  instances.
- No Rust touched, so no `cargo` gate is in scope for this PR.

### Correction I made to my own measurement

I first reported `missing: design-decision` dropping 102 → 92 and treated it as a
possible regression. It was **my grep, not the artifacts**: rivet emits that list
in **non-deterministic order** (`design-decision, feature` on main vs
`feature, design-decision` on the branch for the same artifact), so a literal
string match undercounts. The authoritative section header says
`Lifecycle coverage gaps (102)` → `(103)`, i.e. exactly **+1**, mine. Recording it
because the first number I published was wrong.

## Prediction this PR tests

`Verify-Filter:` is set **in the body at creation time**. #357 set it *after*
`opened`, and I predicted it would not take effect there, because the workflow
reads `github.event.pull_request.body` from the event payload captured at trigger
time and a re-run replays the original payload. This PR is the control: if that
diagnosis is right, the gate here runs 21 artifacts (the stale 20 plus
`TEST-GUARD-HUMAN-SCOPED`) rather than 20. The generic fix — deriving gate scope
from the artifact ids a diff touches — is #358.

Verify-Filter: (and (= type "feature") (or (has-tag "v093") (has-tag "v0100") (has-tag "v0360")))


avrabe:
## The prediction in the body above: **confirmed**

This PR's gate ran the filter set at creation time and matched **21** artifacts, with
`TEST-GUARD-HUMAN-SCOPED` among them — against #357's **20** under the stale default,
where the identical line was added *after* `opened`. From [run 30537212108](https://github.com/pulseengine/spar/actions/runs/30537212108):

```
FILTER: (and (= type "feature") (or (has-tag "v093") (has-tag "v0100") (has-tag "v0360")))
matched 21 artifacts
passed:  21   failed:  0   skipped: 0
```

Two PRs, same filter text, differing only in *when* it entered the body, different scope.
The generic fix is #358, where this is now recorded.

## Merge from `main` — one conflict, and two stale numbers I found in my own artifact

`main` moved (#356, #357). The merge (`457592f`) hit a pure append/append conflict in
`artifacts/verification.yaml` — this branch and #357 each appended a record at the same
position. Both kept verbatim, nothing dropped.

While resolving it I found two wrong numbers in `TEST-GUARD-HUMAN-SCOPED`'s *own*
description. Flagging rather than quietly fixing, because a branch about unbacked claims
not passing quietly does not get an exemption:

- **"two committed fixtures" → three.** The duplicate-id fixture was added after the
  description was written and never enumerated in it, so the artifact under-described its
  own self-test. Now described alongside the other two, with why it exists and the
  mutation it defends against.
- **The "884 committed artifacts" figure is deleted, not corrected.** It was 884 when
  written, 886 when the PR body above quoted it, and **889** after this merge — three
  values in one day. The count isn't load-bearing for the claim it sits in (*zero
  divergence vs PyYAML*), so it's replaced by a re-derived scope description. Correcting
  it would only have produced a slower-moving stale number.

Re-verified after resolution: no conflict markers, both YAML files parse,
`--self-test` **3/3**, real scan **EXIT=0** over 889 artifacts, 3 tagged, all `ok`.

Issue #362, opened 2026-07-30
Trace-Topology nightly has failed 60/60 runs since 2026-06-01 — tools/fixture-vm/flake.lock was never committed


avrabe:
## Summary

`Trace-Topology Fixture Generation` has failed **60 out of its last 60 runs** — every run
back to its oldest recorded one, `2026-06-01`. It has never produced a fixture.

```
$ gh run list --workflow "Trace-Topology Fixture Generation" --limit 60 \
    --json conclusion --jq '[.[]|.conclusion]|group_by(.)|map({(.[0]):length})'
[{"failure":60}]
```

`Fuzz (nightly)` is in the identical state, filed as #361 with a different root cause.

## Cause: a one-time bootstrap that was never done

The workflow's first real step is a deliberate fail-closed guard
(`.github/workflows/trace-fixtures.yml:58`):

```yaml
- name: Require committed flake.lock
  run: |
    if [ ! -f tools/fixture-vm/flake.lock ]; then
      echo "::error::tools/fixture-vm/flake.lock is missing."
      ...
      exit 1
```

And on disk:

```
tools/fixture-vm/
└── flake.nix          # committed 2026-06-11
                       # flake.lock — ABSENT, and not gitignored, just never added
```

**The guard is correct and is working exactly as designed.** The workflow header explains
why it must be this way:

> Determinism: this job builds ONLY against the committed `tools/fixture-vm/flake.lock` —
> it never runs `nix flake update`. The lock IS the pin (the whole reason for the NixOS
> approach); re-resolving inputs per run would silently un-pin nixpkgs / the kernel.

So this is not a broken guard to relax — refusing to build against an unpinned toolchain is
the entire point, and doing it fail-closed is right. The gap is that the companion
bootstrap step was never carried out.

## Fix (the procedure the error message already prescribes)

1. Dispatch `.github/workflows/fixture-vm-flake-lock.yml` (the `fixture-vm flake.lock`
   workflow) — it exists for precisely this.
2. Download its artifact.
3. Commit the resulting `tools/fixture-vm/flake.lock` in a PR, so the pin is reviewed
   rather than resolved silently.

Step 3 needs a human look at what nixpkgs/kernel revision is being pinned — that is a
supply-chain decision, not a mechanical one, so I have not auto-generated it.

## The part that outlives this bug

Same meta-point as #361: neither of these nightlies is a **required** context, so 60
consecutive reds produced no signal anywhere a human looks. A workflow that has never once
[...]

PR #363, opened 2026-07-30
ci(fuzz): stop the scheduler_solver nightly OOMing on ASan bookkeeping (#361)

### LINKED ISSUE #361: Fuzz (nightly) scheduler_solver has failed 97/97 runs since at least 2026-04-25 — ASan allocator amplification, not a solver defect
## Summary

`Fuzz (nightly)` has failed **60 out of its last 60 runs** — every run back to its oldest
recorded one, `2026-06-01`. It has, as far as the run history shows, **never passed.**

```
$ gh run list --workflow "Fuzz (nightly)" --limit 60 --json conclusion \
    --jq '[.[]|.conclusion]|group_by(.)|map({(.[0]):length})'
[{"failure":60}]
```

`Trace-Topology Fixture Generation` is in the identical state and is filed separately.

Nothing surfaced this, because neither workflow is a **required** context — so a
100%-failure gate is indistinguishable on the PR page from a gate that is working. Sixty
consecutive reds produced zero signal. That is the part worth fixing beyond the immediate
cause: **a gate nobody reads is not a gate.**

## The failure

`fuzz_scheduler_solver (1h)` OOMs about 10 minutes in. The other two targets
(`fuzz_aadl_parse`, `fuzz_codegen_roundtrip`) pass.

```
==5922== ERROR: libFuzzer: out-of-memory (used: 2056Mb; limit: 2048Mb)
SUMMARY: libFuzzer: out-of-memory
Error: Fuzz target exited with exit status: 71
```

## It is NOT a defect in `solve_milp` — evidence, not assertion

This was investigated with a two-arm experiment rather than by reading code alone.

**1. The live heap at OOM is 27.7 MB.** From the job's own ASan report:

```
Live Heap Allocations: 27727651 bytes in 8340 chunks;
  quarantined: 155736296 bytes in 2104213 chunks;
  2232955 other chunks; total chunks: 4345508
```

8,340 live chunks after **185,235 executions**. Any per-call leak would put that in the
hundreds of thousands. 4.3M chunks are retained by the *sanitizer* allocator while only
27.7 MB is live.

**2. RSS tracks execution count, not input size.** RSS was already **207 MB while
libFuzzer's max input length was 21 bytes**:

```
#4281    lim: 21    rss: 207Mb
#98083   lim: 580   rss: 1035Mb
#185235  lim: 1240  rss: 2020Mb
```

Throughput collapsed 7772 → 311 exec/s — the signature of a growing allocator working set.

**3. Same workload, ASan on vs off.** A standalone driver building the same MILP shape:

| arm | iterations | RSS | live heap |
|---|---|---|---|
| no ASan | 20,000 | **12.5 MB**, flat after warm-up | — |
| ASan | 120,000 | **582 MB** | **172 KB, constant** from iter ~2,000 on |

**4. Turning down sanitizer bookkeeping alone cuts RSS 3×**, with zero code change:

| `ASAN_OPTIONS` | RSS @ 20k iters |
|---|---|
| defaults | 417 MB |
| `quarantine_size_mb=16` | 187 MB |
| `quarantine_size_mb=16:malloc_context_size=2` | 140 MB |

`solve_milp` is bounded by *counts*, never by *values* — independently re-checked against
the source:

- `milp.rs:67` — `.filter(|(_, t)| t.period_ps > 0)`, so the zero-period
  divide-by-zero hypothesis is **ruled out**, not merely untested.
- `milp.rs:95` — `(0..num_valid * num_procs)` ≤ 8 × 4 = **32** binaries.
- `milp.rs:161` — `vec![0.0_f64; num_procs]`, ≤ 4.
- `memory_bytes` (the only `u32`-wide fuzz field) is **never read** by `solve_milp` — its
  sole occurrence in the file is `milp.rs:247`, inside the test module.

The saved artifact `oom-c5aafda340317c72f5ba9eb2431f0c189d159c9d` (76 bytes) is **not a
reproducer** — it is whichever input happened to be executing when the RSS watchdog fired.
Replaying it will not OOM. Worth stating explicitly so nobody burns a day on it.

Why only this target: it is the only one running a C++ HiGHS MIP presolve + branch-and-bound
per iteration, i.e. thousands of small `new`/`delete` per exec, each becoming a quarantined
ASan chunk.

## Fix

`.github/workflows/fuzz-nightly.yml:62` currently passes only `-max_total_time=3600 -timeout=10`
and sets no `ASAN_OPTIONS`.

1. **`-max_len=128`.** The harness consumes at most 8 tasks (~7 bytes each) + 4 processors
   (~5 bytes) ≈ 80 bytes of entropy; letting libFuzzer grow to 1240-byte inputs is wasted
   work that also inflates per-exec churn. This improves fuzzing *efficiency* as well as memory.
2. **`ASAN_OPTIONS: quarantine_size_mb=32:malloc_context_size=5`** — measured 3× reduction.
   Deep allocation stacks buy little for a target whose only invariant is "must not panic."
3. `-rss_limit_mb=4096` as headroom. Blunt; growth decelerates but does not fully plateau,
   so this buys room rather than a guarantee. Prefer 1 and 2.

## The part that outlives this bug

A nightly that has never once passed should not be able to stay quiet for two months. Worth
deciding: either make these workflows report failure somewhere a human sees (issue-on-fail,
notification), or accept they are advisory and stop presenting them as assurance. Related:
#358, where a gate reported `passed: 20 / failed: 0` over a scope that excluded the very
artifact it was meant to check — same shape, different surface.

Verified against `spar` at `457592f`; workflow line and history re-checked directly.

avrabe:
Closes #361.

## What

`fuzz_scheduler_solver` gets `ASAN_OPTIONS=malloc_context_size=10:quarantine_size_mb=64`
and a per-leg `-max_len=128`. No Rust changes.

## The evidence, not the hypothesis

It has ended in `libFuzzer: out-of-memory` on **every nightly run the API still
returns — 97 of 97**, `2026-04-25` → `2026-07-30`. There is no passing run in the
window at all.

**It is not a leak**, and ASan's own accounting at the moment of death rules the
code under test out ([run 30516991027](https://github.com/pulseengine/spar/actions/runs/30516991027)):

```
used: 2056Mb; limit: 2048Mb
Live Heap:    27.7 MB in     8,340 chunks
quarantined: 155.7 MB in 2,104,213 chunks
total chunks:              4,345,508
```

Live plus quarantined **contents** are 183 MB of 2056 MB. The other **~1.87 GB is
ASan's per-chunk bookkeeping** across 4.3M chunks — a product of two terms, so both
get addressed:

| knob | default | here | term it bounds |
|---|---|---|---|
| `quarantine_size_mb` | 256 | **64** | chunk **count** — 2.1M chunks retained only to catch use-after-free |
| `malloc_context_size` | 30 | **10** | per-chunk **cost** — each stores alloc/free traces of that depth |

Use-after-free is not the bug class this harness hunts; its contract is that
`solve_milp` never panics. Ten frames still names the allocation site, so crash
triage survives — deliberately **not** cut to 2, which saves little and turns a
report into an unactionable address.

`-rss_limit_mb=4096` is headroom, **not the mechanism**. If it turns out to be
load-bearing, the diagnosis above was wrong.

## Why `-max_len` is per-leg — the part worth reviewing

The obvious patch is to append `-max_len=128` to the shared run line. That would
have been a **silent regression**: the line is shared across a 3-target matrix, and
across all 97 runs `fuzz_aadl_parse` and `fuzz_codegen_roundtrip` concluded
`success` **every time**. Only `fuzz_scheduler_solver` fails.

Those two parse arbitrarily long AADL/WIT source text, so a shared cap shrinks two
**healthy** fuzzers' reachable input space to fix a third's bug. Entropy budget is a
property of the harness, so the flag moved into `matrix.include`.

For this leg the cap is **lossless rather than a coverage trade**. The harness opens:

```rust
let n_tasks = input.tasks.len().min(8);
let n_procs = input.processors.len().min(4);
```

while `Arbitrary` grows those `Vec`s to consume whatever buffer it is handed. Every
byte past the ~120 the capped domain can encode yields a `Task` that is allocated
and then discarded unread. Capping removes allocation churn, not reachable states.

## The saved crash artifact is NOT a reproducer

`fuzz/artifacts/fuzz_scheduler_solver/oom-c5aafda340317c72f5ba9eb2431f0c189d159c9d`
is 267 bytes and is simply whichever input happened to be executing when the
process crossed the RSS limit. Replaying it proves nothing. Recording this so
nobody loses a day treating it as the trigger.

## Verification

This is the honest part: **I cannot verify this locally.** No nightly Rust or
`cargo-fuzz` in this environment, and the failure needs a ~10-minute ASan fuzzing
run to reproduce. The numbers above are read from the CI log, and the `-max_len`
argument is read off the harness source; the *fix* itself is so-far unproven.

The oracle is the nightly, and it is **two-sided**:

1. `fuzz_scheduler_solver` must flip `failure` → `success`, **and**
2. `fuzz_aadl_parse` and `fuzz_codegen_roundtrip` must **stay** `success`.

A green run where a parser leg regressed is not a fix. YAML was parsed locally to
confirm the matrix expands to the three intended `(target, extra_args)` pairs.

I will dispatch the workflow on this branch and report the result before merging,
rather than merging on the strength of the reasoning.

## No rivet artifact, deliberately

Matching #353 (the CI path-filter PR), which was also workflow-only and carried
none. There are no `REQ-CI-*` artifacts in this repo — rivet tracks what spar shall
*do*, not CI plumbing. Flagging the omission explicitly so it reads as a choice
rather than an oversight.

## Meta

Neither this workflow nor `trace-fixtures` (#362, 60/60 red) is a **required**
context, which is the whole reason a job that has never once passed sat red for
three months. A gate nobody reads is not a gate. Making them required is a separate
decision — it needs them green first, which is what these two PRs are for.

### DIFF
--- a/.github/workflows/fuzz-nightly.yml
+++ b/.github/workflows/fuzz-nightly.yml
@@ -26,10 +26,26 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target:
-          - fuzz_aadl_parse
-          - fuzz_scheduler_solver
-          - fuzz_codegen_roundtrip
+        # `include` rather than a flat `target:` list because the input-length
+        # cap is a property of the individual harness, not of fuzzing here
+        # (#361). Only fuzz_scheduler_solver fails: across the last 97 runs
+        # the other two legs concluded `success` every time. They consume
+        # arbitrarily long AADL/WIT source text, so a shared -max_len would
+        # shrink the reachable input space of two HEALTHY fuzzers to fix a
+        # third — a silent regression paid for by an unrelated leg's bug.
+        include:
+          - target: fuzz_aadl_parse
+            extra_args: ""
+          # This cap is lossless, not a coverage trade. The harness opens with
+          # `input.tasks.len().min(8)` / `input.processors.len().min(4)`, and
+          # `Arbitrary` grows those Vecs to consume whatever buffer it is
+          # given — so every byte past the ~120 the capped domain can encode
+          # produces a `Task` that is allocated and then discarded unread.
+          # Capping removes allocation churn, not reachable states.
+          - target: fuzz_scheduler_solver
+            extra_args: "-max_len=128"
+          - target: fuzz_codegen_roundtrip
+            extra_args: ""
     steps:
       - uses: actions/checkout@v4
 
@@ -56,10 +72,38 @@ jobs:
       - name: Run fuzz target for 1h
         env:
           TARGET: ${{ matrix.target }}
+          EXTRA_ARGS: ${{ matrix.extra_args }}
+          # This leg has OOMed on every nightly since at least 2026-04-25
+          # (#361). It is NOT a leak in the code under test — ASan's own
+          # accounting at the moment of death says so (run 30516991027):
+          #
+          #   used: 2056Mb; limit: 2048Mb
+          #   Live Heap:    27.7 MB in     8,340 chunks
+          #   quarantined: 155.7 MB in 2,104,213 chunks
+          #   total chunks:              4,345,508
+          #
+          # Live plus quarantined CONTENTS are 183 MB of 2056 MB. The missing
+          # ~1.87 GB is ASan's per-chunk bookkeeping over 4.3M chunks, so the
+          # two knobs below attack the two terms of that product:
+          #   * quarantine_size_mb (default 256) bounds the chunk COUNT — the
+          #     2.1M quarantined chunks are retained solely to catch
+          #     use-after-free, which is not the bug class this harness hunts.
+          #   * malloc_context_size (default 30) bounds the per-chunk COST;
+          #     each chunk stores alloc/free stack traces of that depth.
+          #
+          # 10 frames still names the allocation site, so crash triage keeps
+          # working. Do not cut this to 2 — it saves little and turns a report
+          # into an unactionable address.
+          ASAN_OPTIONS: malloc_context_size=10:quarantine_size_mb=64
         # `--target x86_64-unknown-linux-gnu` avoids the cargo-fuzz default
         # musl target whose statically-linked libc is incompatible with
         # AddressSanitizer. Same fix as the PR-time fuzz smoke (PR #142).
-        run: cargo +nightly fuzz run --target x86_64-unknown-linux-gnu "$TARGET" -- -max_total_time=3600 -timeout=10
+        #
+        # $EXTRA_ARGS is deliberately unquoted: it is a workflow-authored flag
+        # list that must word-split. -rss_limit_mb raises libFuzzer's 2048 MB
+        # default, which the measures above should now keep us under; it is
+        # headroom, not the fix.
+        run: cargo +nightly fuzz run --target x86_64-unknown-linux-gnu "$TARGET" -- -max_total_time=3600 -timeout=10 -rss_limit_mb=4096 $EXTRA_ARGS
 
       - name: Upload corpus
         if: always()


avrabe:
## Oracle satisfied — both sides, full 1h budget

[Run 30560598236](https://github.com/pulseengine/spar/actions/runs/30560598236) — `conclusion: success`.

| target | before | this run |
|---|---|---|
| `fuzz_scheduler_solver` | `failure` (OOM at 2058Mb/2048Mb, 97/97 red since 2026-04-25) | **`success`** |
| `fuzz_aadl_parse` | `success` | `success` |
| `fuzz_codegen_roundtrip` | `success` | `success` |

Both sides of the oracle stated in this PR hold: the OOM leg flipped, and neither parser leg regressed. All three consumed the full 1h budget rather than exiting early, so the green is not a fail-fast artifact.

**Rebased `dc54b6f` → `5199cec`** onto `main` (protection is `strict: true` and #367 had landed). The evidence carries across the rebase: `git diff dc54b6f HEAD -- .github/workflows/fuzz-nightly.yml` is empty, and the only commit main gained was artifacts-only, so nothing under `fuzz/` or the workspace changed between the tree that produced the green run and the tree being merged.

Grounding note: this is a *dispatch* result. The scheduled nightly on `main` remains the real gate and will not have run until this lands.

avrabe:

Verified on `origin/main` before closing, rather than assumed from the batch's commit list: `.github/workflows/fuzz-nightly.yml` carries the ASan-bookkeeping settings (4 matching lines).

Issue #12, opened 2026-03-08
ARINC 653 partition scheduling analysis


avrabe:
## Context
ARINC 653 defines time/space partitioned real-time operating systems used in IMA (Integrated Modular Avionics). OSATE2 has dedicated partition scheduling analysis.

## Requirements
- Partition window scheduling: validate time partition allocations
- Inter-partition communication validation
- Health monitoring configuration validation
- ARINC653 property set support (already partially in standard_properties)

## References
- SAE AS5506/1 Annex C (ARINC653 properties)
- OSATE2: bin packing analysis with ARINC653 support

## Acceptance criteria
- [ ] Partition scheduling validation
- [ ] Inter-partition communication checks
- [ ] ARINC653 property set complete

Issue #11, opened 2026-03-08
WAC composition reading as AADL system implementations


avrabe:
## Context
WAC (WebAssembly Composition) files define how WASM components are wired together. These map directly to AADL system implementations with subcomponents and connections.

## Requirements
- Parse WAC composition format
- Map: component instances → subcomponents, wiring → connections
- Validate compositions against WIT interfaces (already mapped to AADL)
- Generate WAC from AADL system implementations

## Acceptance criteria
- [ ] WAC file produces AADL system implementation ItemTree
- [ ] Subcomponent and connection mapping correct
- [ ] Round-trip: WAC → AADL → WAC preserves semantics

Issue #10, opened 2026-03-08
Rust crate introspection as virtual AADL files


avrabe:
## Context
Rust crates can be modeled as AADL packages, enabling verification that crate interfaces match architectural models.

## Approach
- `cargo metadata --format-version=1` for dependency graph
- `rustdoc --output-format json` for public API (nightly) or source parsing
- Map: crate → package, pub trait → subprogram group, pub fn → subprogram, pub struct → data type

## Requirements
- `spar import --rust <crate-path>` generates virtual AADL from a Rust crate
- Dependency graph becomes with-clause relationships
- Public API mapped to AADL features
- Verification: check that crate implements the interfaces AADL declares

## Acceptance criteria
- [ ] Basic crate → AADL package mapping via cargo metadata
- [ ] Public types/functions mapped to AADL constructs
- [ ] Cross-validation between AADL model and actual crate

PR #364, opened 2026-07-30
ci(fixture-vm): make the nightly actually run — lock, qcow2 attribute, git-dep hash (#362, #365)

avrabe:
Addresses #362 and #365. Neither is auto-closed: a green dispatch is the gate
for flipping those to `Closes`, and that has not happened yet.

> **The scope grew after the first push, deliberately.** This opened as "commit
> the flake.lock", and everything below this box is the *original* text, left
> intact because it was written before I knew the result. It predicted that the
> lock alone would not make the nightly green and that a second failure might be
> hiding behind the first. That was right — twice. Rather than land a lock that
> unblocks nothing and leave the workflow exactly as dead as it was, this PR now
> carries every fix needed for the nightly to actually run.

## What

Adds `tools/fixture-vm/flake.lock`. One new file, 27 lines. Nothing else.

## Why the nightly has never passed

Trace-Topology Fixture Generation has concluded `failure` on **60 of 60 runs**,
`2026-06-01` → `2026-07-30`. There is no passing run in the window.

It is **not a defect** — it is `trace-fixtures.yml:58` doing precisely its job
([run 30516478347](https://github.com/pulseengine/spar/actions/runs/30516478347)):

```
##[error]tools/fixture-vm/flake.lock is missing.
##[error]The nightly build runs only against a committed,
##[error]reviewed lock. Dispatch the 'fixture-vm flake.lock'
##[error]workflow, download its artifact, and commit the
##[error]resulting flake.lock in a PR.
```

The guard is fail-closed by design: the nightly builds only against a committed
lock and never re-resolves inputs itself, so that lock is the determinism pin for
the pinned nixpkgs, the guest kernel, and the whole closure. It has simply never
been satisfied — `tools/fixture-vm/` has contained only `flake.nix` since it
landed. The workflow told us the exact remedy in every one of those 60 runs.

This follows that remedy literally: dispatch
[`fixture-vm flake.lock`](https://github.com/pulseengine/spar/actions/runs/30560238740)
(concluded `success`), download its artifact, commit it in a reviewed PR.

## What is being pinned, and why this commit chooses nothing

`flake.nix:70` declares `github:NixOS/nixpkgs/nixos-24.05`. The lock resolves that
to `b134951a4c9f3c995fd7be05f3243f8ecd65d798`.

I verified that **independently of the container that produced it** — verifying a
CI artifact with the CI that made it proves nothing. Against the GitHub API
directly:

| check | result |
|---|---|
| rev exists in `NixOS/nixpkgs` | ✅ `b134951a…`, `2024-12-30T13:00:28Z` |
| rev **is** the current tip of `nixos-24.05` | ✅ identical sha |

So this commit is the deterministic resolution of what the repo already declares,
not a value I selected.

## ⚠️ For human judgement — deliberately not acted on

**That branch tip is dated 2024-12-30. `nixos-24.05` has not moved in 19 months
and is EOL.**

`nix flake update` faithfully resolved the newest commit on the declared branch;
that branch just stopped receiving them. Bumping to a supported channel is a real
decision and is **not** taken here, for one concrete reason: **it would change the
guest kernel, and the generated trace fixtures are a function of the guest
kernel.** Folding a fixture-changing input bump into a red-to-green CI repair would
hide it.

Two separable questions, and I'd rather they be reviewed as such:

1. *Should the nightly be able to run at all?* — that is this PR.
2. *Should it run on an EOL nixpkgs?* — that is a follow-up, and it needs someone
   who can weigh fixture stability against an unpatched-CVE build environment.

If the answer to (2) is "bump it," the right shape is: change `flake.nix`'s ref,
re-dispatch, and land the fixture diff **visibly** in its own PR.

## Verification — and its honest limit

- Dispatch run `30560238740` — `success`.
- Pinned rev independently confirmed present and branch-tip (table above).
- Lock is well-formed JSON, `version: 7`, single `nixpkgs` input matching
  `flake.nix`.

**What I have not verified:** that the nightly now goes green. This unblocks the
guard at line 58; whether the *subsequent* steps (guest build, `/dev/kvm` boot,
fixture generation) succeed has been untestable for as long as the guard has been
failing — no run has ever reached them. It is entirely possible this reveals a
second failure behind the first. That is progress either way: the next failure
will be a real one rather than a missing file, and it will be visible.

I'll dispatch `trace-fixtures` on this branch and report what actually happens
before merging.

## What the dispatch found: two more blockers, both now fixed here

The prediction above held. The lock let nix evaluate for the first time in the
workflow's life, and evaluation immediately died on something new.

| # | Blocker | Where it surfaced |
|---|---|---|
| 1 | `flake.lock` absent; the line-58 guard is fail-closed | 60/60 runs, never got past it |
| 2 | `system.build.qcow2` never defined — the attribute the workflow builds existed only in comments | run 30560720500, the first run to reach nix |
| 3 | `cargoLock.outputHashes` missing for the `etch` git dependency | local eval, after fixing 2 |

Each was invisible behind the one before it. That is the whole reason this took
60 runs to surface: the outer guard failed closed, so nothing downstream was ever
*reached*, and "never ran" logged identically to "ran and failed."

**Blocker 2** (#365). `system.build` is a freeform `lazyAttrsOf raw`, so a member
nobody defines is not a typo the module system catches — it is
`error: attribute 'qcow2' missing` at evaluation. Fixed by importing nixpkgs' own
`make-disk-image.nix` as a build product, the same way `nixos-generators` defines
its `qcow` format. Two arguments are load-bearing:

- `partitionTableType = "legacy"` — not style. `rootPartition` is a switch on it
  (`legacy → "1"`, `efi → "2"`, `hybrid → "3"`), and `flake.nix` hardcodes
  `fileSystems."/" = "/dev/vda1"`. Any other value partitions cleanly, builds
  cleanly, and then cannot find its root — a failure that appears at guest boot,
  not at build.
- `copyChannel = false` — the default `true` copies the whole nixpkgs tree into an
  image that boots once, runs one oneshot service, and powers off.

**Blocker 3.** `etch` (pulseengine/rivet, pinned in the root `Cargo.toml`) is the
workspace's one git dependency. Registry crates carry their checksum in
`Cargo.lock`; git deps do not, so `importCargoLock` cannot vendor them without an
explicit hash. Note it is vendored even though `gen-fixtures` never uses it —
`importCargoLock` takes the entire lock file, not the subtree `cargoBuildFlags`
selects.

## How this was verified without burning CI runs

`nix build` needs x86_64-linux. But all three blockers are **evaluation** errors,
and evaluation is platform-independent — so the whole guest config can be forced
on a macOS laptop in ~2 minutes, instead of one blocker per ~10-minute self-hosted
dispatch:

```sh
cd tools/fixture-vm
nix eval --raw --no-update-lock-file \
  '.#nixosConfigurations.fixture-vm.config.system.build.qcow2.drvPath'
```

That now exits 0 and prints a `.drv` path, warning-free. Blocker 3 was found this
way rather than by a CI round-trip.

The `etch` hash was cross-checked **two-sidedly against the pinned 24.05
`fetchgit`**, not trusted from the unstable `nix-prefetch-git` that produced it —
a fetcher-behaviour difference between channels would evaluate fine and then fail
at build time in CI. The correct hash yields a store path; a deliberately wrong
one errors with `got: sha256-x37urQw…`, re-deriving the same value independently.
A one-sided check that only confirms the happy path proves nothing.

**What is still unverified:** that the image *builds* and the guest *boots*. Those
are x86_64-linux build-time and run-time concerns that only CI can settle, and a
clean local eval says nothing about either. Dispatched as
[run 30562393849](https://github.com/pulseengine/spar/actions/runs/30562393849); I
will report what actually happens rather than merging on a green evaluation.

Also folded in: `boot.loader.grub.timeout` → `boot.loader.timeout`, which was
emitting a rename warning on every evaluation. One line, and it keeps the build
log readable so a real warning stands out.

## No rivet artifact, deliberately

Matching #353 and #363: workflow/CI-infrastructure changes in this repo carry no
rivet artifacts, and there are no `REQ-CI-*` ids. Flagged so the omission reads as
a choice.

## Meta

Same shape as #361/#363. Neither nightly is a **required** context, which is why a
job that has never once passed sat red for two months announcing its own fix in
plain text. A gate nobody reads is not a gate.


avrabe:
## Verification result — I said I'd dispatch and report what actually happens

I did, and the prediction in the PR body ("it is entirely possible this reveals a
second failure behind the first") **came true**.

[Run 30560720500](https://github.com/pulseengine/spar/actions/runs/30560720500) on
this branch — `conclusion: failure`.

**The good news is real progress.** The fail-closed guard at
`trace-fixtures.yml:58` passed for the first time in 60 runs. nix started,
fetched nixpkgs from `cache.nixos.org`, and got 15 seconds further than any
previous run has ever reached. Then:

```
error: flake 'git+file:///spar?dir=tools/fixture-vm&shallow=1' does not provide attribute
  'packages.x86_64-linux.nixosConfigurations.fixture-vm.config.system.build.qcow2',
  'legacyPackages.x86_64-linux.nixosConfigurations.fixture-vm.config.system.build.qcow2' or
  'nixosConfigurations.fixture-vm.config.system.build.qcow2'
```

`flake.nix` **never defines `system.build.qcow2`**. Every mention of it in the
file is a comment — including the header that documents the exact build command
the workflow runs. `modules = [ … ]` holds exactly one module, and it defines no
`system.build.*`.

Filed with a full diagnosis and a patch sketch checked against the *pinned*
nixpkgs rev: **#365**.

## What I changed here as a result

`Closes #362` → `Addresses #362`. **Merging this will not make the nightly
green**, and a PR that auto-closes the issue would have recorded a fix that
didn't happen. That is the precise failure this repo keeps finding: an operation
that produced "nothing happened" rendering identically to "it worked."

## Why I am still proposing to merge it

Three reasons, and I'd rather they be judged explicitly than assumed:

1. **It is correct on its own terms.** The lock is the deterministic resolution of
   what `flake.nix:70` already declares, verified against upstream independently
   of the container that produced it.
2. **It is what made #365 findable.** Two blockers were stacked and the outer one
   is fail-closed, so all 60 recorded failures died at the guard and the second
   defect had literally never been reached. Removing blocker 1 is what converts an
   opaque red into a specific one.
3. **Reverting costs more than keeping.** Without the lock, the next failure is
   the missing-file error again, and #365 becomes unverifiable.

The alternative — hold this until #365 is also fixed — bundles a reviewed
supply-chain pin with an untested Nix module change into one PR. That is the
bundling #365's scope note argues against.

## Standing caveat, unchanged

Beyond flake evaluation lie the guest build, the `/dev/kvm` boot and fixture
generation. **None of those has ever executed.** I expect more findings behind
#365 and will not describe this pipeline as working until a run reaches the end.

avrabe:
## Dispatch result: it failed. Blocker #4, and it invalidates a claim in this PR.

[Run 30562393849](https://github.com/pulseengine/spar/actions/runs/30562393849) →
`failure`. Reporting it as promised, and correcting myself: the line above that
says this PR "now carries every fix needed for the nightly to actually run" is
**false**. It carries the fixes needed for the nightly to *get as far as
building*, which is where it now dies.

```
building '/nix/store/2bcww30v01dqicgh51s80mwps07y0xbd-etch-0.2.0.drv'...
error: builder for '…etch-0.2.0.drv' failed with exit code 101;
  > failed to load manifest for workspace member `…/rivet-4c06709/rivet-core`
  > Caused by:
  >   feature `edition2024` is required
  >   The package requires the Cargo feature called `edition2024`, but that
  >   feature is not stabilized in this version of Cargo (1.77.1).
```

### The diagnosis, and what it is *not*

This is **not** an `etch` problem, and dropping the dependency would not help.
`etch` is simply the first derivation cargo reached. Every one of spar's 23
crates inherits `edition = "2024"` from `[workspace.package]`, so `gen-fixtures`
itself fails identically a few derivations later.

The real fault is one line at the top of `flake.nix`:

```nix
nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";
```

Measured, not assumed — `rustc.version` at each channel:

| channel | rustc | edition 2024 (needs ≥ 1.85) |
|---|---|---|
| `nixos-24.05` (current pin) | 1.77.2 | ✗ |
| `nixos-24.11` | 1.82.0 | ✗ |
| `nixos-25.05` | 1.86.0 | ✓ |

So the flake has been pinned to a toolchain that **cannot compile this workspace
at all**, for as long as spar has been on edition 2024. Not a regression — this
build has never been possible. Consistent with a workflow that has never once
succeeded.

### Why my local eval did not catch it, which is the point

It couldn't, and I said so before dispatching:

> **What is still unverified:** that the image *builds* and the guest *boots*.
> Those are x86_64-linux build-time and run-time concerns that only CI can
> settle, and a clean local eval says nothing about either.

That boundary held exactly. Evaluation is platform-independent, so blockers 1–3
(all evaluation errors) were findable on a laptop. Blocker 4 is a *compile*
error inside a derivation — on the far side of the line, reachable only by
actually building. The local oracle did its job and stopped precisely where I
said it would stop.

The useful generalisation: **a cheap oracle is worth having only if you state
what it does not cover.** Had I written "verified locally" without the caveat,
this failure would read as the local check being wrong. It wasn't wrong; it was
scoped.

### Running count

| # | Blocker | Class | Found by |
|---|---|---|---|
| 1 | `flake.lock` absent; guard fail-closed | workflow | 60 red runs |
| 2 | `system.build.qcow2` never defined | evaluation | first run to reach nix |
| 3 | `cargoLock.outputHashes` missing for `etch` | evaluation | local eval |
| 4 | nixpkgs pin's Rust predates edition 2024 | **build** | this dispatch |

Four deep, each invisible behind the last. Pushing the pin bump next; the
`etch` `outputHashes` value must be re-verified against 25.05's `fetchgit`
rather than carried over, since a fetcher-behaviour change between channels
would evaluate fine and fail at build time — the same trap, one layer down.

I am not flipping "Addresses" to "Closes" and not merging until a dispatch is
actually green.

avrabe:
## Dispatch result: failed. Blocker #5, and it falsifies a claim I put in this branch.

Reporting the actual outcome as committed, not merging on a green evaluation.

[Run 30563612749](https://github.com/pulseengine/spar/actions/runs/30563612749)
**failed** at `Build guest + generate fixtures`. It got further than any of the
64 runs before it — nix evaluated, the whole NixOS closure built, the Rust
dependency tree built — and then died in `gen-fixtures` itself:

```
error: rustc 1.86.0 is not supported by the following package:
  smol_str@0.3.6 requires rustc 1.89
```

### This makes a sentence in `1ea73b1` false

That commit moved the pin 24.05 → 25.05 and wrote into `flake.nix`:

```
    nixos-25.05 → 1.86.0   ✓
```

with `✓` defined two lines above as "can compile this workspace". **It cannot.**
I am flagging that explicitly rather than quietly correcting it, because the way
it went wrong is worth more than the fix.

Edition 2024 requiring ≥ 1.85 is a **necessary** condition that I mistook for
**the** condition. The probe genuinely ran and genuinely returned `1.86.0` — so
the check *felt* like verification, while silently scoping the claim to the one
variable I had happened to think of. A one-variable check that comes back *true*
is the most convincing way to not verify something.

The real constraint is

```
max(rust-version) over the entire RESOLVED DEPENDENCY CLOSURE
```

which is an emergent property of `Cargo.lock`, **not** of this repo. The
workspace declares no `rust-version` and carries no `rust-toolchain.toml`, so
nothing here states that floor and nothing here pins it. Derived mechanically
this time instead of guessed a second time — 152 of 235 packages declare one:

```sh
cargo metadata --format-version 1 --locked \
  | jq -r '.packages[] | select(.rust_version) | .rust_version' \
  | sort -V | tail -1
# 1.89   (smol_str 0.3.6; next is 1.87.0 — wasip2, wit-bindgen)
```

| channel | rustc | verdict |
|---|---|---|
| nixos-24.05 | 1.77.2 | ✗ edition 2024 not stabilised |
| nixos-24.11 | 1.82.0 | ✗ likewise |
| nixos-25.05 | 1.86.0 | ✗ clears the edition, **fails the closure** |
| nixos-25.11 | 1.91.1 | ✓ clears both |

Pushed as `aef79c2`: pin → `nixos-25.11`.

### Verified against the instantiated derivation, not `pkgs.rustc`

Same discipline as blocker #4 — the question is what the *toolchain reaching
gen-fixtures* is, and `pkgs.rustc` is not that. The `gen-fixtures-0.10.0.drv`
reachable from the disk image now references `cargo-1.91.1.drv` and
`rustc-wrapper-1.91.1.drv`.

The `etch` `outputHash` was **re-derived** under the new channel rather than
assumed to survive it — a `fetchgit`-defaults change between channels evaluates
fine and fails only at build time. Checked from the side that *cannot pass
vacuously*: a deliberately wrong hash forces a real fetch and reports

```
got:    sha256-x37urQw97R/ARqvlVpXpp3tJqbvztbOiUyAGNZItlA0=
```

identical to the pinned value. The happy path proves nothing here — a
fixed-output derivation short-circuits when the store path already exists, so
"it built" would have been true whether or not the hash was still correct.

### The blocker ladder so far

| # | Blocker | Surfaced |
|---|---|---|
| 1 | `flake.lock` absent; the guard is fail-closed | 60/60 runs, never got past it |
| 2 | `system.build.qcow2` never defined | run 30560720500, first run to reach nix |
| 3 | `cargoLock.outputHashes` missing for the `etch` git dep | local eval |
| 4 | pinned channel's cargo predates edition 2024 | run 30562393849 |
| 5 | closure MSRV 1.89 > channel's rustc 1.86 | run 30563612749 |

Each was invisible behind the one before it. Five deep is not bad luck — it is
what a fail-closed outer guard does to a pipeline nobody has ever run to
completion. Every one of these has been sitting there since the workflow was
written; none of them is a regression.

### Known-unfixed, stated so it is not mistaken for coverage

The closure floor **moves on every `cargo update`**. A dependency bump can raise
it with no change to this repo and no change to this flake, and **nothing gates
that** — it surfaces only as a red nightly here, at build time, which is exactly
how blocker #5 got in. The comment in `flake.nix` therefore carries the
*derivation command* rather than a number to trust. A real fix is either a
declared `rust-version` in `[workspace.package]` or a CI check comparing the two;
both are out of scope for this PR and I would rather name the gap than imply it
is closed.

Re-dispatched as [run 30564693866](https://github.com/pulseengine/spar/actions/runs/30564693866).
Still **`Addresses`, not `Closes`**, and still not merging until a dispatch is
actually green.

avrabe:
## Dispatch result: further than ever, still red — blockers #6 and #7

Reporting [run 30564693866](https://github.com/pulseengine/spar/actions/runs/30564693866)
as committed. It **failed**, and it is the most informative failure this
workflow has produced.

The 25.11 pin worked. Cargo accepted the closure (no MSRV complaint), every
vendored crate compiled, and the build reached `gen-fixtures-0.10.0.drv` itself
before dying:

```
error: builder for '/nix/store/…-gen-fixtures-0.10.0.drv' failed with exit code 101
  >   failed to execute command: No such file or directory (os error 2)
  >   is `cmake` not installed?
```

**Blocker #6.** `highs-sys` compiles the HiGHS C++ LP solver from vendored
source via the `cmake` crate, and `nativeBuildInputs` had only `pkg-config`. It
sits five levels below anything this image is for:

```
spar-trace-topology → spar-network → good_lp → highs → highs-sys
```

Note `gen-fixtures` never calls a solver. `--bin gen-fixtures` selects what gets
**linked**, not what gets **compiled** — cargo builds the dependency graph of the
whole `-p spar-trace-topology` package. Exactly the same shape as the `etch`
vendoring already documented in this flake: the lock file decides what must
build, not the binary.

`dontUseCmakeConfigure = true` goes with it and is load-bearing, not defensive.
Putting cmake in `nativeBuildInputs` makes nixpkgs' setup-hook install
`cmakeConfigurePhase` as this package's `configurePhase` — it guards on exactly
that variable at `setup-hook.sh:145`, which I checked in the pinned tree rather
than recalling — and it would then try to cmake a Rust workspace root that has
no `CMakeLists.txt`.

## Blocker #7 was predicted, not discovered

The thing that has made this PR expensive is that **each blocker was invisible
behind the one before it**, so every dispatch bought exactly one bit of
information. Six dispatches, six bits. So instead of pushing the cmake fix and
waiting, I read the dependency graph looking for the next one:

```
$ cargo tree -p spar-trace-topology -i clang-sys
clang-sys v1.8.1
└── bindgen v0.71.1
    [build-dependencies]
    └── highs-sys v1.12.1
```

`highs-sys` also build-depends on **bindgen**, which reaches libclang through
`clang-sys → libloading` — it `dlopen()`s `libclang.so` while the build script
runs. Nothing in the sandbox provides that, so `Unable to find libclang` was
queued up to surface the moment the cmake fix cleared the way, costing another
~10-minute round trip to learn one more bit.

`rustPlatform.bindgenHook` exports `LIBCLANG_PATH` and `BINDGEN_EXTRA_CLANG_ARGS`
and nothing else here sets them. Both it and the cmake hook's guard variable
were confirmed present in the pinned nixpkgs source tree, not assumed from
nixpkgs conventions.

`clang-sys` and `highs-sys` are the only two `-sys` crates in the graph, so this
should close the native-dependency species. "Should" is doing real work in that
sentence.

## Running tally

| # | Blocker | How it was found |
|---|---------|------------------|
| 1 | `flake.lock` absent, fail-closed guard | 60/60 runs |
| 2 | `system.build.qcow2` never defined | first run to reach nix |
| 3 | `cargoLock.outputHashes` missing for `etch` | local eval |
| 4 | nixpkgs rustc predates edition 2024 | local eval |
| 5 | closure MSRV 1.89 > 25.05's rustc 1.86 | CI dispatch — **falsified my own claim in this PR** |
| 6 | `cmake` missing for highs-sys | CI dispatch |
| 7 | `libclang` missing for highs-sys' bindgen | **read the graph, before dispatching** |

## Still unverified

Evaluation is clean, and evaluation is platform-independent — it says nothing
about whether the image builds or the guest boots. A clean eval has now been
wrong about this build three times. Dispatched as
[run 30565973744](https://github.com/pulseengine/spar/actions/runs/30565973744).

Unchanged commitment: still **`Addresses`**, not `Closes`, and **still not
merging until a dispatch is actually green.**

avrabe:
## Dispatch result: the Rust half is done — blocker #8 is the disk image

[Run 30565973744](https://github.com/pulseengine/spar/actions/runs/30565973744)
failed, as committed. Blockers #6 and #7 are **closed**: `gen-fixtures-0.10.0.drv`
built successfully and took 3m22s doing it, so cmake + libclang were the whole
of that problem. The Rust closure now compiles end-to-end inside the image.

It died one layer further out, in the disk image itself:

```
error: a 'x86_64-linux' with features {kvm} is required to build
'/nix/store/…-nixos-disk-image.drv', but I am a 'x86_64-linux' with features
{benchmark, big-parallel, nixos-test, uid-range}
```

**Blocker #8.** That feature set is Nix's default *minus* `kvm`. Nix adds `kvm`
only when `access("/dev/kvm", R_OK|W_OK)` succeeds (`settings.cc`,
`getDefaultSystemFeatures`), and `make-disk-image.nix` builds through
`vmTools.runInLinuxVM`, whose derivation carries
`requiredSystemFeatures = [ "kvm" ]`.

The device was already being passed. `--device /dev/kvm` was in the podman
invocation from the start, and the fleet comment is accurate — the runner user
*is* in the `kvm` group on the host. The gap is that **rootless podman drops the
caller's supplementary groups by default**. `/dev/kvm` is `crw-rw---- root:kvm`,
so inside the user namespace the process belongs to no group that can open it.
The node was present and unopenable — which is why `--device` looked sufficient
and wasn't. `--group-add keep-groups` carries them through.

## The diagnostic existed; nobody could reach it

Nix's error is correct, precise, and names the exact missing feature. It was
emitted after ~5 minutes of compiling, as line ~700, under a solid wall of
`building '/nix/store/…drv'`. I only found it by tailing `--log-failed`.

So this commit also adds a precondition step that asks Nix for its computed
`system-features` **before anything is built** — same fact, two seconds in, at
the top of the log. It asserts on Nix's own feature set rather than a shell
`[ -w /dev/kvm ]` test, because the feature set is what the build is actually
matched against; a probe of anything else could green while the build still
fails. The image digest moves to workflow `env:` so the probe and the build it
guards cannot drift onto different images.

**Status of the new dispatch
([30567928227](https://github.com/pulseengine/spar/actions/runs/30567928227)):**
`Require /dev/kvm usable inside the container` — **success**. So
`keep-groups` works on this runtime and Nix now advertises `kvm`. The qcow2
build is running as I write this.

## Running tally

| # | blocker | closed by |
|---|---|---|
| 1 | nightly refuses to run without a committed `flake.lock` | f18be8d |
| 2 | `system.build.qcow2` is not a stock NixOS option | 1f5ef9a |
| 3 | `etch` git dependency hash | 1f5ef9a |
| 4 | nixpkgs 24.05's cargo predates edition 2024 | 1ea73b1 |
| 5 | closure MSRV is 1.89, above 25.05's rustc 1.86 | aef79c2 |
| 6 | `highs-sys` needs `cmake` | a33d0d6 |
| 7 | `highs-sys` needs `libclang` | a33d0d6 |
| 8 | rootless podman drops the `kvm` group | 52fafa0 |

Eight blockers, each invisible until the one before it cleared — the cost of a
job that has never once run to completion. Every fix so far has been real and
none of them were guessable in advance.

Still `Addresses`, not `Closes`, and **still not merging until a dispatch is
actually green**. Blocker #8 being diagnosed is not blocker #8 being fixed; the
qcow2 step has never completed, and there are four steps after it — boot the
guest, tshark the capture, round-trip all four fixtures — that have never
executed at all. I will report the outcome either way.

avrabe:
## Blocker #8 is closed — measured, not inferred. Blocker #9 is open.

Reporting the outcome as committed. Run
[30567928227](https://github.com/pulseengine/spar/actions/runs/30567928227):
**`Build NixOS fixture-vm qcow2` succeeded** — 13 minutes, ending at
`building '/nix/store/b6lxg7a95bbsz5x66g2vpx9isbfq7w9y-nixos-disk-image.drv'`.
That is the first time this workflow has ever produced a disk image.

I said I would read the probe's log rather than infer "keep-groups worked" from
"the step exited 0". Its own output:

```
container process: uid=0(root) gid=0(root) groups=0(root),65534(nobody)
crw-rw---- 1 nobody nobody 10, 232 Jul 28 04:16 /dev/kvm
nix system-features = benchmark big-parallel kvm nixos-test uid-range
ok: nix advertises kvm; nixos-disk-image.drv is buildable here
```

The host `kvm` group arrives inside the user namespace as `65534`, which is also
what `/dev/kvm`'s group maps to there — so the process can open it, and Nix's own
`getDefaultSystemFeatures()` `access(/dev/kvm, R_OK|W_OK)` probe consequently puts
`kvm` in the feature set. That supplementary group is precisely what rootless
podman drops and what `--group-add keep-groups` carries through.

## Blocker #9 — `Locate qcow2`

The next step failed, and it looks like a naming mismatch. It is not one.

```
nix build … --out-link /spar/tools/fixture-vm/result
```

wrote the out-link **into the bind-mounted repo**, pointing at `/nix/store/…`.
The container is `--rm` and carries its own store; after the run the symlink
survives on the host and its target does not. The build product was unreachable
from the host **by construction** — globbing for a differently-named file would
have found nothing, because the problem was never the name.

Fixed in dccc117: the out-link goes to `/tmp` and the image is copied across the
mount while the store it lives in is still alive.

`Locate qcow2` also stops trusting `cp`'s exit code. A copy that exits 0 says a
file arrived, not that a bootable NixOS image did — a truncated or wrong-path
copy would have surfaced three steps later at `Run fixture-vm`, far from its
cause. It now asserts existence plus a ≥ 64 MiB floor and names what went wrong.

Also: neither `nixos.qcow2` nor `result` was ever gitignored. That stayed
harmless only because the build had never once got far enough to produce them.

## Tally

| # | blocker | closed by |
|---|---|---|
| 1 | nightly refuses to run without a committed `flake.lock` | f18be8d |
| 2 | `system.build.qcow2` is not a stock NixOS option | 1f5ef9a |
| 3 | `etch` git dependency hash | 1f5ef9a |
| 4 | nixpkgs 24.05's cargo predates edition 2024 | 1ea73b1 |
| 5 | closure MSRV is 1.89, above 25.05's rustc 1.86 | aef79c2 |
| 6 | `highs-sys` needs `cmake` | a33d0d6 |
| 7 | `highs-sys` needs `libclang` | a33d0d6 |
| 8 | rootless podman drops the `kvm` group | 52fafa0 ✅ *verified green* |
| 9 | the qcow2 never leaves the container's store | dccc117 |

Dispatched [30569595760](https://github.com/pulseengine/spar/actions/runs/30569595760).

Still `Addresses`, not `Closes`, and **still not merging until a dispatch is
actually green**. Four steps have still never executed: boot the guest, tshark
the capture, round-trip all four fixtures, upload. Each is a fresh opportunity
for blocker #10. I will report the outcome either way.

avrabe:
## Reporting the outcome, as committed: the dispatch was **not** green.

Run [30569595760](https://github.com/pulseengine/spar/actions/runs/30569595760)
**failed**. Honest summary first, diagnosis after.

### What passed — including new ground

| # | step | result |
|---|---|---|
| 3 | Require committed flake.lock | ✅ |
| 4 | Require /dev/kvm usable inside the container | ✅ `crw-rw---- 10, 232` |
| 5 | Build NixOS fixture-vm qcow2 | ✅ (~14 min) |
| 6 | Locate qcow2 | ✅ |
| 9 | Build fixture-vm harness | ✅ |
| **11** | **Run fixture-vm (boot NixOS guest)** | ❌ **failed after 900s** |
| 12–14 | tshark / round-trip / upload | ⏭ skipped |

The **qcow2 byte count I said I'd read back** once logs were servable:

```
fixture-vm: qcow2 = …/tools/fixture-vm/nixos.qcow2 (1995309056 bytes)
-rw-r--r-- 1 runner9 runners 1.9G Jul 30 20:31 …/nixos.qcow2
```

**1,995,309,056 bytes — 1.9 GiB**, about 30× the 64 MiB floor that step
enforces. The image is a real NixOS system, not a truncated copy. Blocker #9
(the container-store copy) stays closed.

Step 11 is a step that had **never executed** in this workflow's history. So
the guest genuinely booted for the first time — and then this:

```
18:34:57  fixture-vm: booting QEMU guest …
18:34:58  SeaBIOS (version 1.16.3-debian-1.16.3-2)
18:34:58  iPXE (https://ipxe.org) 00:03.0 …
18:34:58  Booting from Hard Disk...
18:34:58  GRUB loading.
18:34:58  Welcome to GRUB!
          ← 900 seconds of nothing —
18:49:57  fixture-vm: error: QEMU guest did not shut down within 900 seconds; killed
```

### The silence is not the guest hanging at GRUB — it is where we go blind

The obvious read is "the guest wedged right after GRUB." I don't think that's
what happened, and the reason is specific.

The host side is correct and was never the problem: the harness runs
`-nographic -serial mon:stdio` with `Stdio::inherit()`
(`crates/spar-trace-topology/src/bin/fixture-vm.rs:198`). What survived it is
only **firmware** output. Under `-nographic`, SeaBIOS redirects the **BIOS text
console** onto the serial port; GRUB draws through BIOS calls, so GRUB rode that
redirect for free.

Linux does not use BIOS calls. At handoff it switches to whatever `console=`
names — and the default is `tty0`, the VGA framebuffer `-nographic` has
disconnected. `tools/fixture-vm/flake.nix` sets `boot.kernelPackages`,
`boot.kernelModules` and `boot.loader.grub`, but **never `boot.kernelParams`**.
So from kernel handoff onward the kernel, systemd, and `gen-fixtures` have all
been logging into a void — and `serial-getty@ttyS0`, which the flake *does*
enable, attaches to a `/dev/console` that isn't the serial port either.

"Welcome to GRUB!" is not where the guest stopped. It is the last line we were
**able to see**.

### The fix (8f581aa)

```nix
boot.kernelParams = [ "console=tty0" "console=ttyS0,115200" ];
```

Order is load-bearing: every `console=` receives kernel messages, but the
**last** one becomes `/dev/console`, which is what init and the getty attach to.
`ttyS0` must come last or the getty lands back on the framebuffer.

### What this does and does not claim

It does **not** claim the guest now completes. I have no evidence for that and
won't assert it. It makes the *next* failure legible instead of a 900-second
blank — which is the difference between debugging and guessing.

That is the same shape as every other blocker in this series, and worth naming
because it keeps recurring: **an operation that produced "nothing happened"
rendered identically to one that worked.** #365 was an attribute that existed
only in prose; #9 was a `cp` that exited 0 into a store that no longer existed;
this is a console that was configured on the host and never on the guest.

Still not merging until a dispatch is actually green. Re-dispatching on 8f581aa
now, and I'll report that outcome either way too.

avrabe:
Dispatch outcome on `8f581aa`: **failed.** Run [30572572863](https://github.com/pulseengine/spar/actions/runs/30572572863) — reporting as promised, still not merging.

It failed *earlier* than the commit was written for. `8f581aa` addressed step 11 (`Run fixture-vm`, the 900-second blind boot); this run died at step 5, **`Build NixOS fixture-vm qcow2`**. So the console fix was never exercised — it is neither confirmed nor refuted by this run.

## What actually happened

`nixos-disk-image.drv` exited **134** (SIGABRT). Two lines carry it:

```
qemu-system-x86_64: Could not access KVM kernel module: Permission denied
qemu-system-x86_64: failed to initialize kvm: Permission denied
qemu-system-x86_64: falling back to tcg
...
qemu-system-x86_64: ../system/physmem.c:758: iotlb_to_section:
    Assertion `section_index < d->map.sections_nb' failed.
```

`make-disk-image.nix` builds the qcow2 by booting a transient VM. That VM could not open `/dev/kvm`, fell back to software emulation, and TCG then aborted on a QEMU assertion at `-smp cpus=32`. The abort is a consequence; the EPERM is the cause.

## Why the KVM precondition passed anyway

This is the part worth writing down, because the precondition step was added specifically to stop this class of failure, and it reported green two minutes earlier:

```
container process: uid=0(root) gid=0(root) groups=0(root),65534(nobody)
crw-rw---- 1 nobody nobody 10, 232 /dev/kvm
nix system-features = benchmark big-parallel kvm nixos-test uid-range
```

Both readings are correct. `/dev/kvm` is mode `0660`, group = the host `kvm` gid, which is unmapped in the rootless userns and so renders as `65534`/`nobody`. The container process is uid 0 **with supplementary group 65534** — `--group-add keep-groups` worked. So `access("/dev/kvm", R_OK|W_OK)` succeeds, Nix adds `kvm` to `system-features`, and the derivation is legally schedulable here. All true.

The build still cannot open the device, because **the process that answered the probe is not the process that runs the build.** Nix computes `system-features` in the parent (uid 0, groups `0,65534`); it then runs the builder as a `nixbld` build user, which carries no such supplementary group. Root may open `/dev/kvm`; nixbld may not.

The step's own comment defends the design:

> It asserts on Nix's OWN computed feature set rather than on a shell `[ -w /dev/kvm ]` test: the feature set is what the build is actually matched against, so a probe of anything else could pass while the build still fails.

That is right about **scheduling** and blind to **execution**. `system-features` governs whether the derivation may be *placed* on this machine. The failure was in whether its builder may *open the device*. Different principal, different syscall, different question — and the more rigorous-looking probe asks the one that was not at stake. A `[ -w /dev/kvm ]` test would have been equally wrong, for the same reason: both run as the wrong user.

Same shape as every other blocker in this series: an operation that produced "nothing happened" — here, "no KVM for you" — rendered identically to one that worked.

## What I am changing

1. **Probe by doing the thing.** Replace the config read with a real derivation, `requiredSystemFeatures = ["kvm"]`, whose builder is `/bin/sh -c 'exec 3<>/dev/kvm'`. That runs as the *build* principal and performs the *actual* `open(O_RDWR)`. It costs seconds and needs no nixpkgs. It also prints `id` from inside the builder, which is the line that would have made this obvious immediately instead of 12 minutes into a 700-line build log.
2. **Make the probe container and the build container share one nix.conf.** They currently differ — the build appends `experimental-features` and `sandbox = false`; the probe appends nothing. This workflow already hoisted `NIX_IMAGE_DIGEST` into `env:` on exactly this reasoning ("a probe that greens a different image than the one built proves nothing about the build"); a probe with different Nix settings than the build has the same defect.
3. **The leading fix: run builds as the container root** (`build-users-group =`), so the builder keeps the `keep-groups` supplementary group. I am labelling this a hypothesis, not a finding — I have confirmed root can open the device and the builder cannot, but I have not yet confirmed the image sets `build-users-group = nixbld`. The new probe settles it either way, in seconds, and prints `nix config show build-users-group` alongside.

Note that TCG is not a usable fallback here regardless — it aborts. So KVM has to genuinely work in the builder; failing loudly at the top is the only correct behaviour.

I will report the next dispatch outcome either way, as before.

avrabe:
Follow-up: the `build-users-group` item is no longer a hypothesis. I said I would settle it, and it settles from the image's own definition rather than from waiting on the run — [NixOS/nix@2.24.9 `docker.nix`](https://github.com/NixOS/nix/blob/2.24.9/docker.nix):

```nix
nixbld${toString n} = { uid = 30000 + n; gid = 30000; groups = [ "nixbld" ]; };

groups = { root.gid = 0; nixbld.gid = 30000; nobody.gid = 65534; };

defaultNixConf = { sandbox = "false"; build-users-group = "nixbld"; … };
```

Three things fall out.

**1. The diagnosis is confirmed.** The image ships `build-users-group = "nixbld"`, so builds drop to uid 30001+. Each nixbld user's supplementary group list is exactly `[ "nixbld" ]` — gid 30000, never 65534. `/dev/kvm` in the container is group 65534. So the builder cannot open it, while the parent (uid 0, carrying 65534 via `--group-add keep-groups`) can. That is the whole failure, and `build-users-group =` is the right fix.

**2. The daemon caveat is refuted.** I flagged that if the image ran a nix *daemon*, appending to `/etc/nix/nix.conf` after startup would not reach the builder and the fix would be inert. There is no `NIX_REMOTE` and no daemon in the image — `nix` runs standalone as root and re-reads `/etc/nix/nix.conf` on each invocation. The override will take effect.

**3. An incidental correction.** `sandbox = "false"` is *already* the image default, so this workflow's appending of it has been redundant since it was written. Harmless, but the comment above it implies it is doing work it is not. I will fold that into the same follow-up.

It also explains a cosmetic detail from the failing run's log that looked like noise:

```
crw-rw---- 1 nobody nobody 10, 232 /dev/kvm
```

`nobody` there is not a real owner — the host's `root` and `kvm` ids are unmapped in the rootless userns, so both fall through to the overflow id 65534, which this image happens to name `nobody`. Two different unmapped ids printing the same name is exactly the kind of rendering that made "present but unopenable" look like "present and fine".

Run 30575158798 is still queued behind the single podman runner, and I will report it either way. But note what its job now is: it is no longer the experiment that decides the diagnosis — that is decided. It is the confirmation that the fix works end to end, and the first chance the boot fix in `8f581aa` gets to run at all.

Still not merging until a dispatch is actually green.

avrabe:
## Reporting run 30575158798, as promised — it **failed**

I said I would report it either way and would not merge until a dispatch is
actually green. It is not green, so **this PR stays open**.

`conclusion: failure`. Queued 19:31:02 → started 20:07:16 (**36m14s** waiting for
runner-9, the only `podman` host — see #380) → failed 20:16:53.

### What the run does establish

| step | result |
|---|---|
| 3. Require committed flake.lock | success |
| **4. Require /dev/kvm openable by a Nix BUILD process** | **success** |
| 5. Build NixOS fixture-vm qcow2 | **failure** |

Step 4 is the fix in `62610b5`, and it did the thing it was built to do:

```
container process: uid=0(root) gid=0(root) groups=0(root),65534(nobody)
crw-rw---- 1 nobody nobody 10, 232 Jul 30 18:49 /dev/kvm
nix build-users-group = []
nix system-features   = benchmark big-parallel kvm nixos-test uid-range
building '/nix/store/ygpsx197vb02an2dw3ad209w78q9c1wb-kvm-open-probe.drv'...
ok: a Nix BUILD process opened /dev/kvm O_RDWR.
```

Step 5 then ran for **5m21s** and built several hundred derivations — the whole
NixOS unit graph, then the Rust closure through `wit-bindgen`, `wasmparser`,
`cargo-vendor-dir`, and into `gen-fixtures-0.10.0.drv`. The previous run died at
this step in seconds with `Could not access KVM kernel module: Permission
denied`. So the KVM class of failure is gone, and this is a *new* wall.

`nix build-users-group = []` also empirically confirms what the (still unpushed)
comment commit `56a9b26` asserts: the empty override is the load-bearing line,
not the `sandbox = false` that the old comment credited.

### Why it failed — blocker #12, and it is caused by `sandbox = false`

```
error: home directory '/homeless-shelter' exists; please remove it to assure
purity of builds without sandboxing
```

With the sandbox off, Nix runs every builder with `HOME=/homeless-shelter` and
refuses to start a build if that path actually **exists** — an existing home
directory would let builders read state from outside the derivation. Some
builder earlier in the graph created it (a `cargo`/`rustc` build phase doing the
usual `mkdir -p "$HOME/.cargo"` is the standard culprit), and the next
derivation's purity check then refused to run.

Note the coupling: this check *only exists* when the sandbox is off, which is
the image default we are relying on. It is the direct downstream cost of that
choice, so the fix has to be chosen with that in mind rather than papered over
with an `rm -rf` in the wrapper — a pre-step removal cannot help when the
directory is created *mid-graph*.

### A defect in the probe itself, found by the same log

```
kvm-open-probe> sh: line 2: id: command not found
kvm-open-probe> builder principal:
```

`kvm-probe.nix` runs `echo "builder principal: $(id)"` so that, per its own
comment, "it prints `id` first, so the log names that principal outright."
It printed **nothing**. Nix clears `PATH` for builders and this derivation is
deliberately dependency-free, so there is no `coreutils` and therefore no `id`.
`echo` still succeeded, so `set -eu` did not catch it.

That is exactly the failure this probe was *written to fix*, one level in: a line
whose whole job was to be evidence rendered "nothing happened" identically to
"it worked". The **assertion** is unaffected — `exec 3<>/dev/kvm` succeeded and
the derivation produced its output, so a Nix build process really did open
/dev/kvm O_RDWR — but the identification did not survive its own environment.

Fix is to read the principal with shell built-ins only, and to fail if it comes
back empty rather than print a bare label:

```sh
uid=""
while read -r k v; do case "$k" in Uid:) uid="$v" ;; esac; done < /proc/self/status
[ -n "$uid" ] || { echo "cannot identify the builder principal"; exit 1; }
echo "builder principal Uid: $uid"
```

### Next

Both fixes go on this branch and I re-dispatch. Not merging on a red dispatch.

avrabe:
## Dispatch re-run: red. Not merging. Here is what moved and what did not.

Run [30579194737](https://github.com/pulseengine/spar/actions/runs/30579194737) on `b57200b`
— `failure`, 20:28:20 → 20:39:31 UTC.

I said on this PR: *"Both fixes go on this branch and I re-dispatch. Not merging on
a red dispatch."* It ended up being three fixes, not two, and the dispatch is red.
So: not merging. Correcting my own count while I am here.

### What the three fixes actually bought

| step | result |
|---|---|
| 3 `Require committed flake.lock` | success |
| 4 `Require /dev/kvm openable by a Nix BUILD process` | **success** |
| 5 `Build NixOS fixture-vm qcow2` | failure |

Each fix is discharged by evidence, not by the run being green:

- **`77e66b3` (keep gen-fixtures out of `/homeless-shelter`) — worked.**
  `gen-fixtures-0.10.0.drv` started 20:32:25 and finished 20:38:05: five minutes
  forty, completed. In run 30575158798 that same derivation died 5m21s in with
  `home directory /homeless-shelter exists`. The `preConfigure` HOME redirect is
  the difference.
- **`4baeb47` (probe the real build principal) — worked.** Step 4 is green, and
  green now means an actual `open(3<>/dev/kvm)` from inside a derivation with
  `requiredSystemFeatures = ["kvm"]`, with the uid read from `/proc/self/status`
  by shell built-ins and a hard `exit 1` if it comes back empty.
- **`b57200b` (name the culprit before the container evaporates) — worked, and it
  is worth saying how.** It printed:

  ```
  --- nix build exited 1; impure-HOME diagnostic ---
  /homeless-shelter is absent, so this is NOT the impure-
  HOME failure -- read the build log above instead.
  ```

  That is the *negative* branch. A diagnostic that only speaks when it is the
  culprit reads identically to a diagnostic that never ran — which is the precise
  bug `4baeb47` fixed one level down, where `builder principal:` printed empty
  because `id` did not exist. Ruling a hypothesis out is the output, not the
  silence.

The failure moved roughly six minutes deeper into the build graph. That is the
measurement; it is not a green run.

### The new failure is a different one

`nixos-disk-image.drv` fails. Two `virtiofsd` instances die identically:

```
ERROR virtiofsd] can't apply the child capabilities: failed to sync capabilities with the kernel
```

and then, inevitably:

```
qemu-system-x86_64: -chardev socket,id=store,path=virtio-store.sock: Failed to connect to 'virtio-store.sock': Connection refused
```

The socket refuses because nothing ever bound it. `virtiofsd` is what serves the
Nix store into the builder VM; it exits during sandbox setup, so qemu has nothing
to connect to. The `Connection refused` is the symptom, three lines below the cause.

This is the **same species** as fixture-vm blockers #8 and #11, one axis over.
Both of those were about what a rootless container hands the build: #8 was the
supplementary **group** (`--group-add keep-groups`), #11 was which **principal**
the probe ran as. This one is the **capability** set. The container gets
`--device /dev/kvm --group-add keep-groups --security-opt label=disable` and
nothing about capabilities, so it runs with podman's rootless default set —
which omits `SYS_ADMIN`, `MKNOD` and `DAC_READ_SEARCH`, and `virtiofsd`'s
sandbox needs all three plus `SETPCAP` to drop the bounding set afterwards.
`capng_apply` failing on a bounding-set change is exactly what "failed to sync
capabilities with the kernel" is.

### What happens next

The fix goes on this branch, with a probe that asserts the capabilities are
present rather than a flag added on faith — same discipline as step 4, so that
deleting the flag turns the probe red instead of turning the build red eleven
minutes later.

Nothing is blocked by holding this open: `trace-fixtures` is **not** one of the
18 required contexts on `main` (checked against
`branches/main/protection/required_status_checks`), so this PR gates only itself.

PR #368, opened 2026-07-30
fix(rivet): stop double-loading safety/stpa and de-duplicate RENDER-REQ (#360)

### LINKED ISSUE #360: Required rivet gate cannot detect duplicate artifact ids — 6 real ones on main, masked by 184 self-inflicted phantoms
## Summary

The required `Rivet validate (artifacts)` gate **cannot detect a duplicate artifact
id**, and `main` currently has six real ones. Six requirement records in
`artifacts/requirements.yaml` — the primary artifact file — are **silently ignored
by rivet**, and two of them state a different status than the one actually in force.

Found while building the `human-scoped` guardrail (#359), whose duplicate-id fixture
was written against this. Filed separately because choosing which copy is canonical
is a content decision about the rendering requirements, not a guardrail change.

## Evidence: the gate is green and the tool that finds this is pinned out

CI's pinned **rivet v0.4.3**, green run
[30525896163](https://github.com/pulseengine/spar/actions/runs/30525896163) on `main`:

```
Lifecycle coverage gaps (102):

Result: PASS (127 warnings)
```

`ERROR:` lines: **0**. Lines matching `declared more than once`: **0**.

Local **rivet 0.28.0** on the same tree exits **1** with **466 errors**. Exhaustive
accounting — the two classes sum to exactly 466, nothing unexplained:

| class | count |
|---|---|
| `status '<x>' is not an allowed value` | 276 |
| `artifact id '<x>' is declared more than once` | 190 |

## Defect A — 184 of those 190 duplicates are self-inflicted by `rivet.yaml`

`rivet.yaml` lists the whole `safety/stpa` directory **and** four individual files
*inside* it:

```yaml
sources:
  - path: artifacts
    format: generic-yaml
  - path: safety/stpa            # <- scans the directory
    format: stpa-yaml
  - path: safety/stpa/requirements.yaml        # <- and again, individually
    format: generic-yaml
  - path: safety/stpa/architecture.yaml
    format: generic-yaml
  - path: safety/stpa/validation.yaml
    format: generic-yaml
  - path: safety/stpa/solver-requirements.yaml
    format: generic-yaml
```

So each of those four files is loaded **twice** and every id in it collides with
itself. That is why the message reads `./safety/stpa/validation.yaml and
./safety/stpa/validation.yaml` — same path on both sides.

Verified there are **no** within-file duplicates on disk:

| file | `- id:` lines | unique | dup within file | rivet errors |
|---|---|---|---|---|
| `safety/stpa/validation.yaml` | 66 | 66 | **0** | 66 |
| `safety/stpa/requirements.yaml` | 51 | 51 | **0** | 51 |
| `safety/stpa/solver-requirements.yaml` | 35 | 35 | **0** | 35 |
| `safety/stpa/architecture.yaml` | 32 | 32 | **0** | 32 |
| `safety/stpa/rendering-analysis.yaml` | 15 | 15 | **0** | 6 (cross-file — see B) |

184 phantom errors, one per id in the four double-listed files.

**This is the load-bearing part.** `ci.yml:395` justifies pinning the gate to v0.4.3
because newer rivet produces *"464 phantom errors vs the pinned v0.4.3's genuine
PASS"*. That label is right about 184 duplicates and wrong about 6 — so the pin
suppresses real findings as collateral of suppressing self-inflicted noise. Fixing
the config is what makes the pin decision re-evaluable on merit.

Not a trivial deletion: the four files are listed explicitly to get
`format: generic-yaml`, while the directory scan applies `stpa-yaml`. Dropping the
explicit entries changes how they are parsed. Options: move generic artifacts out of
`safety/stpa/`, or ask rivet for a per-file format override / exclude within a
directory source.

## Defect B — RENDER-REQ-001…006 are genuinely duplicated, and the copies contradict

Each id is defined **twice**, in two separately-scoped files, with drifted content:

| id | `artifacts/requirements.yaml` | `safety/stpa/rendering-analysis.yaml` | in force |
|---|---|---|---|
| RENDER-REQ-003 | `implemented` | `partial` | **`partial`** |
| RENDER-REQ-006 | `partial` | `implemented` | **`implemented`** |

Contradicting each other in opposite directions. Titles diverge too
(`Port visibility with directional indicators` vs `Visible ports with directional
indicators`; `Semantic zoom for overview levels` vs `Semantic zoom`).

**The `safety/stpa` copy wins** — confirmed from the v0.4.3 CI log, which reports
`hazards` / `satisfies` / `verified-by` fields on all six. Only the `safety/stpa`
copies carry those; the `artifacts/requirements.yaml` copies are bare stubs with no
trace links at all.

Consequences:

1. **Six records in `artifacts/requirements.yaml` are dead.** A human or agent
   greps the primary artifact file, reads `RENDER-REQ-003: implemented`, and is
   wrong — rivet evaluates `partial`. The losing record is not flagged; it just
   stops existing.
2. **`RENDER-REQ-003`'s honest status is the one being masked in the file people
   read.** The `safety/stpa` copy explains itself — *"Phase 3a (zoom/pan/selection)
   is implemented; Phase 3b (minimap/search) is deferred"* — while
   `artifacts/requirements.yaml` flatly claims `implemented`.
3. **Which copy wins looks hash-ordered, not deterministic.** v0.4.3 emits the six
   ids as 001, 005, 003, 002, 004, 006. Same root cause as the non-deterministic
   `missing: <types>` list ordering (filed as rivet friction). Today `safety/stpa`
   happens to load second; nothing in the repo pins that.

## Also surfaced: two status spellings for one concept

Of the 276 status-value errors, most are schema-declaration skew (the repo's
lifecycle vocabulary for test artifacts isn't declared in the 0.28.0 schema). But
nine are genuine internal inconsistency:

| value | count |
|---|---|
| `passing` | 171 |
| `not-implemented` | 51 |
| `planned` | 37 |
| `partial` | 8 |
| **`pass`** | **5** |
| **`pending`** | **4** |

`pass` (5) alongside `passing` (171), and `pending` (4) alongside `planned` (37) —
two spellings each for the same state, which no gate currently catches.

## Proposed fix, in order

1. **Deduplicate `rivet.yaml` sources** so the 184 phantom errors disappear. This
   must come first: it is what makes the remaining error count meaningful.
2. **Pick the canonical RENDER-REQ copy** — recommend keeping `safety/stpa/
   rendering-analysis.yaml` (it holds the trace links and the honest, reasoned
   statuses) and deleting the six stubs from `artifacts/requirements.yaml`. Needs a
   decision on whether losing them from the primary file is acceptable.
3. **Normalize `pass`→`passing` and `pending`→`planned`.**
4. **Then re-evaluate the v0.4.3 pin on merit** — with 1 and 3 done, the residual
   0.28.0 error count should be near zero, and the pin can be lifted so duplicate
   ids become detectable by the required gate rather than only by a local run.

Steps 1–3 are mechanical and independently landable. Step 4 is the one that turns
this from a cleanup into a gate that would have caught it.

avrabe:
> [!WARNING]
> **Two claims below were refuted by cold-context verification and are withdrawn.** See [the correction comment](https://github.com/pulseengine/spar/pull/368#issuecomment-5134269008) and #371.
>
> 1. **"the directory scan alone loses nothing"** — measured on `rivet list --format json` *without* `--full`, a view that omits `fields` entirely. Re-measured: 5 artifacts (VAL-010/011/014/015/021) lose half their `test-name`. Fixed in `cce6fbd`; filed upstream as pulseengine/rivet#747.
> 2. **"the repo cannot widen `allowed-values` locally"** and its corollaries **"the v0.4.3 pin is load-bearing"** / **"measured as not viable"** — false. Vendoring `schemas/common.yaml` (v0.3.0, same as embedded) and widening the vocabulary takes `rivet validate` 276 → 0, exit 0. The remedy is still declined, but on the merits (#371), not on a constraint that doesn't exist.
>
> The duplicate-id result — 190 → 6 → **0** — is unaffected and was independently confirmed exact at all four measurement rows.

---

Closes #360.

`rivet validate` reported **466 errors on a clean tree**. This takes it to 276,
eliminates the duplicate-id class entirely, and — where the issue's own plan was
wrong — says so with the measurement rather than quietly dropping that step.

## The 190 phantom duplicates

`rivet.yaml` listed `safety/stpa` as a directory scan **and** listed four files
inside it individually, to force `format: generic-yaml` on them. Each of those
files was therefore loaded twice, and every id in them was "declared more than
once" — **colliding with itself**, with the same path on both sides of the
message.

Listing a directory and its own contents isn't additive, it's duplicative, and
the diagnostic couldn't hint at that because both copies had identical
provenance.

**Equivalence was measured, not assumed.** Dropping the explicit entries changes
those files' parse format, so before/after `rivet list --format json` was
compared: **883 ids and 1195 links, byte-identical**. `stpa-yaml` is a superset
of what `generic-yaml` was extracting here.

## The 6 real ones

`RENDER-REQ-001..006` were defined in **both** `artifacts/requirements.yaml` and
`safety/stpa/rendering-analysis.yaml`. rivet kept one copy and dropped the other
**with no diagnostic**, and which one won was load order, not a decision.

I deleted the `requirements.yaml` stubs and kept the STPA copy, because only the
STPA file carries the `hazards`/`satisfies`/`verified-by` links that make these
part of the STPA V. The stubs carried none.

**The stubs were stale in both directions**, which is why I checked against
source instead of keeping whichever file looked fresher:

| id | stub said | actually | evidence |
|----|-----------|----------|----------|
| RENDER-REQ-003 | `implemented` | `partial` | `minimap` is `/// Show minimap (Phase 3b — reserved)` |
| RENDER-REQ-006 | `partial` | `implemented` | `etch/src/html.rs:65-68` emits `svg.zoom-low` / `svg.zoom-overview`; `:131` asserts a `wheel` handler |

Verified at the **pinned etch rev `4c06709`**, not by grepping this repo — and
that distinction nearly cost me a false finding. A repo-scoped grep shows
`zoom-low` only in a checked-in playwright fixture and `semantic zoom` only in a
README bullet, which reads exactly like "asserted against a hand-authored
artifact, never implemented." It reads that way because the implementation lives
in a **git dependency**. "Not in this repo" and "not implemented" are
indistinguishable to grep.

## Where the issue was wrong

#360 predicted that with the overlap and the vocabulary both fixed, "the
residual 0.28.0 error count should be near zero." **That is false, and this PR
measures it rather than asserting it:**

| step | errors | duplicate-id |
|------|--------|--------------|
| baseline | 466 | 190 |
| drop source overlap | 282 | 6 |
| delete RENDER-REQ stubs | 276 | **0** |
| normalise `pass`/`pending` | **276** | 0 |

The normalisation **cannot** reduce the count, because its targets aren't in the
allowed set either:

```
allowed: draft, proposed, approved, implemented, verified,
         released, accepted, deprecated, rejected
```

Neither `passing` nor `planned` is there. So `pass` → `passing` moves 5 errors
between buckets (172 → 177) and `pending` → `planned` moves 4 (36 → 40). Net
zero, exactly as measured.

It's still worth doing — it leaves **4** non-conforming values to resolve instead
of 6 — but it is not an error fix, and booking it as one would have overstated
what this change buys.

## Why the v0.4.3 pin still can't be lifted

The residual 276 is entirely status-vocabulary skew: 177 `passing`, 51
`not-implemented`, 40 `planned`, 8 `partial`. These are **test/verification**
states, not approval-lifecycle states — a VAL record's "passing" is about whether
the test passes right now, which is a genuinely different axis from where an
artifact sits in an approval flow.

rivet names the two options itself: fix the artifacts, or add the values to the
`status` base-field's `allowed-values`. **The repo can't take the second one** —
`common@0.3.0` is embedded in the rivet binary, and `schemas/` only overrides
`sysml2`.

So `ci.yml`'s pin to `rivet-cli v0.4.3` is **load-bearing for the green check**,
not incidental. Lifting it turns a required context red with 276 errors that are
real under the current schema. The existing comment there calls a newer rivet's
findings "phantom errors"; for the duplicate-id class that was true and this PR
fixes it, but for the status class it isn't — those are genuine
non-conformances against `common@0.3.0` that the pin is holding back.

Resolving them is an upstream rivet conversation (does the lifecycle admit
test-execution states?), not something to force in this PR.

## What this does not do

- Does not reduce the 276 status errors. Named above with the reason.
- Does not lift the v0.4.3 pin. Measured as not viable; see above.
- Does not touch `safety/stpa/rendering-analysis.yaml`. The canonical statuses
  there were independently confirmed correct against etch, so there was nothing
  to change.

### DIFF
--- a/artifacts/requirements.yaml
+++ b/artifacts/requirements.yaml
@@ -260,54 +260,19 @@ artifacts:
     tags: [integration, serde]
 
   # ── Rendering (STPA-derived) ──────────────────────────────────────────
-
-  - id: RENDER-REQ-001
-    type: requirement
-    title: Orthogonal edge routing
-    description: >
-      Edges must use orthogonal routing to minimize visual crossings.
-    status: implemented
-    tags: [rendering, stpa]
-
-  - id: RENDER-REQ-002
-    type: requirement
-    title: Port visibility with directional indicators
-    description: >
-      Ports must be visible with directional indicators and type coloring.
-    status: implemented
-    tags: [rendering, stpa]
-
-  - id: RENDER-REQ-003
-    type: requirement
-    title: Interactive HTML with zoom/pan/minimap/search
-    description: >
-      Interactive HTML output must support zoom, pan, minimap, and search.
-    status: implemented
-    tags: [rendering, stpa]
-
-  - id: RENDER-REQ-004
-    type: requirement
-    title: Deterministic layout
-    description: >
-      Layout must be deterministic — same model always produces same layout.
-    status: implemented
-    tags: [rendering, stpa]
-
-  - id: RENDER-REQ-005
-    type: requirement
-    title: Selection and group highlighting
-    description: >
-      Selection and group highlighting must be supported in interactive mode.
-    status: implemented
-    tags: [rendering, stpa]
-
-  - id: RENDER-REQ-006
-    type: requirement
-    title: Semantic zoom for overview levels
-    description: >
-      Semantic zoom must reduce clutter at overview levels.
-    status: partial
-    tags: [rendering, stpa]
+  #
+  # RENDER-REQ-001..006 are NOT defined here. They live in
+  # safety/stpa/rendering-analysis.yaml, which is the canonical copy: it
+  # carries the hazards/satisfies/verified-by trace links that make them
+  # part of the STPA V, and the stubs that used to sit here carried none.
+  #
+  # Both files defined all six ids until #360. rivet silently kept ONE and
+  # dropped the other with no diagnostic, and which one won was load order,
+  # not a decision. The stubs were stale in BOTH directions -- they claimed
+  # RENDER-REQ-003 `implemented` (it is `partial`: minimap/search are
+  # Phase 3b, still `reserved` in etch) and RENDER-REQ-006 `partial` (it is
+  # `implemented`: etch/src/html.rs emits svg.zoom-low / svg.zoom-overview
+  # behind a wheel handler). Do not re-add them here.
 
   - id: REQ-WASM-001
     type: requirement
--- a/artifacts/verification.yaml
+++ b/artifacts/verification.yaml
@@ -732,7 +732,7 @@ artifacts:
 
   - id: VAL-CODEGEN-001
     type: feature
-    status: pass
+    status: passing
     title: Golden model integration tests for codegen
     description: >
       19 golden model integration tests validating end-to-end code generation
@@ -754,7 +754,7 @@ artifacts:
 
   - id: VAL-CODEGEN-002
     type: feature
-    status: pass
+    status: passing
     title: WIT/Rust/Config/Test/Proof generation output validation
     description: >
       Tests validating each codegen output layer independently: WIT interface
@@ -795,7 +795,7 @@ artifacts:
 
   - id: VAL-SYSML2-LOWER-001
     type: feature
-    status: pass
+    status: passing
     title: SysML v2 lowering tests
     description: >
       Tests for SysML v2 to AADL lowering covering diagnostic handling for
@@ -814,7 +814,7 @@ artifacts:
 
   - id: VAL-VERIFY-MACROS-001
     type: feature
-    status: pass
+    status: passing
     title: Verify macro proc-macro tests
     description: >
       Tests for spar-verify-macros proc-macro crate covering #[aadl(...)]
@@ -837,7 +837,7 @@ artifacts:
 
   - id: VAL-KANI-SOLVER-001
     type: feature
-    status: pending
+    status: planned
     title: Kani — solver output implies no deadline miss
     description: >
       kani_schedule_implies_no_deadline_miss harness: for any TaskSet of
@@ -861,7 +861,7 @@ artifacts:
 
   - id: VAL-KANI-SOLVER-002
     type: feature
-    status: pending
+    status: planned
     title: Kani — priority ordering is monotone and antisymmetric
     description: >
       kani_priority_monotonicity harness: for any two tasks in a ≤4-task
@@ -885,7 +885,7 @@ artifacts:
 
   - id: VAL-KANI-SOLVER-003
     type: feature
-    status: pending
+    status: planned
     title: Kani — EDF accepts U=1 at n=2, RM rejects
     description: >
       kani_zero_laxity_handled harness: a 2-task implicit-deadline set
@@ -909,7 +909,7 @@ artifacts:
 
   - id: VAL-KANI-CODEGEN-001
     type: feature
-    status: pending
+    status: planned
     title: Kani — codegen emission preserves schedule task IDs
     description: >
       kani_emit_preserves_schedule harness: for any Schedule of size
@@ -980,7 +980,7 @@ artifacts:
 
   - id: VAL-MUTATION-001
     type: feature
-    status: pass
+    status: passing
     title: Mutation testing for analysis passes
     description: >
       Mutation testing campaign covering 674 analysis tests. Validates that
--- a/rivet.yaml
+++ b/rivet.yaml
@@ -18,16 +18,16 @@ project:
 sources:
   - path: artifacts
     format: generic-yaml
+  # The directory scan below already loads every *.yaml under safety/stpa.
+  # Four of those files used to ALSO be listed individually, to force
+  # `format: generic-yaml` on them — which loaded each one twice and made
+  # every id in it collide with itself: 184 phantom "declared more than
+  # once" errors, one per id, with the same path on both sides of the
+  # message (#360). The explicit entries are removed; see the measurement
+  # in the PR for the artifact-set equivalence check that says the
+  # directory scan alone loses nothing.
   - path: safety/stpa
     format: stpa-yaml
-  - path: safety/stpa/requirements.yaml
-    format: generic-yaml
-  - path: safety/stpa/architecture.yaml
-    format: generic-yaml
-  - path: safety/stpa/validation.yaml
-    format: generic-yaml
-  - path: safety/stpa/solver-requirements.yaml
-    format: generic-yaml
 
 docs:
   - docs
--- a/rivet.yaml
+++ b/rivet.yaml
@@ -23,9 +23,28 @@ sources:
   # `format: generic-yaml` on them — which loaded each one twice and made
   # every id in it collide with itself: 184 phantom "declared more than
   # once" errors, one per id, with the same path on both sides of the
-  # message (#360). The explicit entries are removed; see the measurement
-  # in the PR for the artifact-set equivalence check that says the
-  # directory scan alone loses nothing.
+  # message (#360). The explicit entries are removed.
+  #
+  # This is NOT free, and the PR's first measurement said it was — wrongly.
+  # `rivet list --format json` WITHOUT `--full` emits only the summary view
+  # (`--help`: it omits `description`, `tags`, and `fields`), so the
+  # "byte-identical, loses nothing" equivalence check was run on a view
+  # structurally incapable of seeing field-level loss — the exact class of
+  # change a `format:` switch produces. Re-measured with `--full`:
+  #
+  #   +86 artifacts GAIN `mitigates` and `traces-to` (stpa-yaml really is
+  #       broadly a superset — but the original measurement didn't show that
+  #       either, it showed nothing)
+  #    -5 artifacts LOSE half of `test-name`: VAL-010/011/014/015/021 are
+  #       exactly the five whose on-disk value is a comma-separated pair, and
+  #       the stpa-yaml parser keeps only the first element.
+  #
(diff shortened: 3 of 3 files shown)


avrabe:
## Correction — two of this PR's claims were wrong, found by cold-context adversarial verification

I ran a fresh-context agent against this PR's own claims with no inherited framing. It refuted two, and I re-derived both myself before acting. Pushed `cce6fbd` for the one that was baked into a committed file; the other is corrected here and in #371.

### 1. "The directory scan alone loses nothing" — FALSE, and it was committed into `rivet.yaml`

The equivalence check I cited ran `rivet list --format json` **without `--full`**. Per `list --help`, that emits only the summary view — it omits `description`, `tags`, and `fields`. So the "byte-identical" result was measured on a view **structurally incapable of detecting field-level loss**, which is exactly the class of change a `format:` switch produces. The measurement was true and meaningless.

Re-measured with `--full`, base vs. this branch:

- **+86** artifacts gain `mitigates` and `traces-to` — `stpa-yaml` really is broadly a superset, but my original measurement didn't show that either
- **−5** artifacts lose half of `test-name`

The five are VAL-010, VAL-011, VAL-014, VAL-015, VAL-021 — **exactly** the five whose on-disk `test-name` is a comma-separated pair (`grep -c 'test-name:.*,' safety/stpa/validation.yaml` → `5`). A/B on identical bytes, only `format:` differing:

| `format:` | `rivet get VAL-010 --format json \| jq -r '.fields["test-name"]'` |
|---|---|
| `generic-yaml` | `tests::boolean_matches_aadl_boolean, tests::boolean_mismatches_aadl_string` |
| `stpa-yaml` | `tests::boolean_matches_aadl_boolean` |

`test-name` is *evidence* — the pointer from a verification artifact to the tests discharging it. The trade is still worth taking (5 truncated pointers against 190 phantom duplicate-id errors), but it is a **trade, not a free win**, and the committed comment now says so. Filed upstream as pulseengine/rivet#747.

### 2. "The repo cannot widen `allowed-values` locally" — FALSE

I claimed the residual 276 errors were unfixable here because `common@0.3.0` is embedded in the binary. rivet's own output says otherwise, verbatim:

```
common  0.3.0  embedded  (compiled into rivet — changes on upgrade; vendor under schemas/ to pin)
```

Measured:

```
$ cp <rivet>/schemas/common.yaml schemas/common.yaml    # version "0.3.0" — identical to embedded
$ rivet validate | tail -2
Result: FAIL (276 errors, 629 warnings, 0 broken cross-refs)
Schemas: common@0.3.0 (on-disk), ...                     # override live

$ # append passing / not-implemented / planned / partial to status allowed-values
$ rivet validate | tail -2 ; echo exit=$?
Result: PASS (629 warnings)
exit=0
```

**276 → 0.** My claim inferred *"only `sysml2` **can** be on-disk"* from the observation *"only `sysml2` **is** on-disk"* — an inference reported as a measurement. That's the same one-variable-check failure this repo keeps re-learning.

**Consequently these two statements in the PR body are withdrawn:** *"`ci.yml`'s pin to `rivet-cli v0.4.3` is load-bearing for the green check"* and *"measured as not viable."* The pin is load-bearing *today*, but for a fixable reason.

### …and the remedy should still be declined — now on the merits

Now that the option is known to exist, #371 declines it deliberately rather than on a false constraint. Widening the vocabulary on a vendored fork of the org-shared `common` schema would silence a **correct** diagnostic: rivet already ships `verification-verdict` — *"Pass/fail verdict for a single verification measure in an execution run"* — and spar is recording verdicts in the lifecycle `status` field of `type: feature` (*"a user-visible capability"*). The 276 errors are a real modeling defect, not a vocabulary gap. Softening the ledger to green it is the textbook anti-pattern.

### Smaller corrections, for the record

- **"rivet kept one copy and dropped the other with no diagnostic"** — inaccurate at 0.28.0, which names both files *and* the winner (`— the second definition silently overwrites the first`). True only of the pinned v0.4.3.
- **"Which one won was load order, not a decision"** — load order *is* a decision: it's the `sources:` order in `rivet.yaml`, and `Store::upsert` is documented last-write-wins. Fully deterministic. #360's "looks hash-ordered" is refuted; I softened but inherited it.
- **The repo-grep anecdote is overstated.** `zoom-low` also appears in `tests/playwright/rendering.spec.ts:145-147` — an executable assertion, not a fixture — and `semantic zoom` appears in `artifacts/verification.yaml:474`, the same spec, and a docs plan. The rhetorical point stands; the grep offered as evidence for it does not.
- **Adjacent defect this PR canonises without checking:** keeping the STPA copies makes `RENDER-REQ-006 verified-by: [VAL-R5]` authoritative, and VAL-R5's evidence is `run: cargo test -p spar-wasm -- semantic_zoom` — which matches **zero** tests repo-wide. `playwright` appears in no workflow. The status is correct against etch source; its verification pointer is dead. Tracking separately.

The core of the PR is unaffected — the duplicate-id class still goes 190 → 6 → **0**, and that measurement was independently confirmed exact at all four rows.

avrabe:

Verified on `origin/main` before closing, rather than assumed from the batch's commit list: `rivet.yaml` has the de-duplicated `safety/stpa` sources (2 entries, not the doubled set).

Issue #370, opened 2026-07-30
release: field is on the wrong plane for 59 of 85 artifacts — rivet's release-planning query is blind to 11 shipped releases


avrabe:
## Summary

The repo uses **two different conventions for the `release:` field, and rivet honors
only one of them.** 26 artifacts declare it at artifact top level (rivet's core field
plane) and 59 declare it nested under `fields:` (rivet's custom-extension bag). The
release-planning query surface — `rivet list --release <VERSION>`, the mechanism the
methodology says drives the roadmap ("readiness is a query, not an opinion") — is
blind to the 59.

**11 shipped releases report zero artifacts to rivet:** v0.13.0, v0.14.0, v0.15.0,
v0.16.0, v0.17.0, v0.18.0, v0.19.0, v0.20.0, v0.21.0, v0.31.0, v0.32.0.

This is the release-planning analogue of #360: a query that silently reports less
than the truth, where *"no artifacts in this release"* is indistinguishable from
*"the field is in the wrong place."*

## Measurement (origin/main @ 2b11af9)

```
$ grep -c '^    release: ' artifacts/requirements.yaml      # top-level, 4-space
26
$ grep -rc '^      release: ' artifacts/ safety/            # fields:-nested, 6-space
artifacts/requirements.yaml:58
safety/stpa/rendering-analysis.yaml:1
                                                            # -> 59

$ rivet list --format json | jq '[.artifacts[]|select(.release != "-")]|length'
25          # matches the top-level grep on that checkout exactly
```

`rivet list` grouped by release: **856 of 881 artifacts report `release: "-"`.**

## Mechanism (proven, not inferred)

rivet has two field planes:

- **core fields** (`id`, `type`, `status`, `tags`, `release`) — read by the CLI's own
  query surface, including `rivet list --release`, documented as *"Filter to artifacts
  scoped to a release — the release-planning view. Sugar for `(= release "<ver>")`"*.
- **the `fields:` extension bag** — schema-declared custom keys. For `requirement` the
  declared set is `[baseline, category, cited-source, priority, upstream-ref]`.

Putting `release:` in the bag is syntactically valid, schema-tolerated, and invisible
to the query. Confirmed by exact set correlation:

```
ids rivet INFO-flags for undeclared 'release':  59
ids with fields:-nested release:                58  (+1 in rendering-analysis.yaml)
INFO set MINUS nested set:                      RENDER-REQ-006   (the rendering-analysis one)
[...]

Issue #294, opened 2026-06-16
friction: emv2_propagation reports 0 chains for all real models — AADL .emv2 annex not ingested into the propagation overlay


avrabe:
**What I was trying to do:** Get EMV2 fault-propagation chains to analyse for `spar/relay_transport.aadl` — the forge/replay/jam link threats reaching the channel's AEAD/anti-replay sinks — to mechanize the STPA-Sec mitigation story (SH-FORGE/SH-STALE/SH-JAM ↔ SSC-AUTH/SSC-FRESH/SSC-JAM) as an AADL fault-propagation counterpart.

**Tool + version:** Unknown command: --version

**What happened:** `spar analyze` always reports

```
emv2_propagation: EMV2 propagation: 0 chain(s), 0 local flow(s) analysed
```

for **any** real AADL model, regardless of how the EMV2 error flows are declared. Root cause is documented in the source — `crates/spar-analysis/src/emv2_propagation.rs:128`:

> // Because EMV2 annex data is not yet integrated into `SystemInstance`,
> // callers supply this overlay when constructing analysis inputs.
> // Production code will eventually populate this from the parsed `.emv2`
> // annex subclause; in the interim, tests build it programmatically.

So the AADL `annex EMV2 {** ... **}` (`error source`/`sink`/`path` + `propagation`) is parsed for the **fault tree** and the **EMV2→STPA bridge** (both work — they report sources/SPFs/hazards), but it is **not fed into the `Emv2Overlay`** that `compute_error_propagation` consumes. On the `spar analyze` path that overlay is empty, so the propagation analysis is effectively a no-op for real models.

**Repro (two independent models, both 0 chains):**
```
# 1. relay's transport model (error sources on the bus)
spar analyze --root Relay_Transport::SecureTransport.Link spar/relay_transport.aadl

# 2. spar's OWN bundled OSATE example, using the canonical bindings/connection
#    bus-binding idiom — also 0 chains:
spar analyze --root network_protocol_pkg::MySystem.basic \
  test-data/interop/osate-examples/LargeExamplesforEMV2TR/network_protocol_pkg.aadl
```
I also tried moving the sources to port out-propagations (`error source frame_out {...}` on a port that IS in a semantic connection to the sink port) — still 0 chains, consistent with the overlay never being populated from the annex.

[...]

Issue #372, opened 2026-07-30
Two CI jobs run ~25 min and cannot fail the build; two real gates were unrequired


avrabe:
## Summary

Two CI jobs run on every code PR, consume ~25 minutes of runner time between them,
and are **structurally incapable of failing the build**. A third pair of genuinely
load-bearing jobs was running unrequired until today.

Found by a cold-context audit of the release machinery (not by a failure — which is
the point: a gate that can't go red never announces itself).

## 1. `Kani Bounded Model Checking` — 23m01s, cannot go red

`.github/workflows/ci.yml:456` — `continue-on-error: true`, at job level.

The file documents the intent to fix this at lines 491-497:

> `continue-on-error: true` — these harnesses are first-pass; unwind […] flip
> continue-on-error to false and add to the required-checks list.

That hasn't happened. Today the job proves nothing enforceable: a Kani harness can
regress to `FAILED` and every PR still merges green.

## 2. `Bazel test (//...)` — no-op *and* `continue-on-error: true`

`.github/workflows/ci.yml:508`. Doubly vacuous — even if the `continue-on-error`
were removed, there is nothing to build:

```
$ git ls-tree origin/main --name-only | grep -iE 'MODULE.bazel|WORKSPACE'
(no output)
```

The job's own script `ls`-es for the workspace markers, finds none, prints a warning
and `exit 0`. It is a cache-warming step wearing a test job's name — and the name is
the problem, because "Bazel test (//...)" reads like coverage in the checks list.

## 3. (fixed) Two real gates were not required

`Codegen compile oracle` and `Verification Gate (rivet-driven)` both run on every PR,
neither has `continue-on-error`, and both can genuinely go red — yet neither was in
`required_status_checks.contexts` (16 entries, verified via
`gh api .../branches/main/protection`).

`codegen-oracle` is the **only** thing executing the `#[ignore]`d compile oracles plus
three runtime oracles (`exec-oracle`, `state-oracle`, `kiln-oracle` — the wasmtime and
kiln data-flow proofs for REQ-CODEGEN-WIT-DATAPLANE-001 / -STATE-001 / -KILN-001).
`cargo nextest` does not run `#[ignore]` tests, so **nothing required covered that
ground**: a PR breaking generated-component codegen, the WIT ABI round-trip, or the
spar→meld→kiln chain went red there and merged anyway.

`verification-gate.yml`'s final step `Fail job if any verification artifact failed`
[...]

PR #374, opened 2026-07-30
fix(artifacts): move 276 statuses onto rivet's lifecycle vocabulary

### LINKED ISSUE #371: Verification verdicts are recorded as feature.status — the root cause of all 276 rivet validate errors
## Summary

`artifacts/verification.yaml` records **pass/fail test-execution verdicts** in the
`status:` field of `type: feature` artifacts. `status` is rivet's *lifecycle* field
(`draft → proposed → approved → implemented → verified → released`) and `feature` is
*"a user-visible capability or feature"*. Neither models a test result.

That mismatch is the entire content of the 276 `rivet validate` errors currently
outstanding:

```
177  status 'passing'         is not an allowed value
 51  status 'not-implemented' is not an allowed value
 40  status 'planned'         is not an allowed value
  8  status 'partial'         is not an allowed value
--- 
276
```

rivet already ships the right types (`rivet schema list`):

```
verification-execution    A verification execution run against a specific version
verification-verdict      Pass/fail verdict for a single verification measure in an execution run
```

## Correcting the record from #360 / #368

PR #368 (and my comment on #360) asserted that the repo **cannot** take either remedy
rivet's help names, because `common@0.3.0` is embedded in the binary. **That is false,
and I've measured it false.** rivet's own `schema sources` output says so verbatim:

```
common  0.3.0  embedded  (compiled into rivet — changes on upgrade; vendor under schemas/ to pin)
```

Vendoring it works, and widening the vocabulary makes the errors vanish:

```
$ cp <rivet>/schemas/common.yaml schemas/common.yaml     # version 0.3.0 — same as embedded
$ rivet validate | tail -2
Result: FAIL (276 errors, 629 warnings, 0 broken cross-refs)
Schemas: common@0.3.0 (on-disk), ...                      # <- override is live

$ # append passing / not-implemented / planned / partial to status allowed-values
$ rivet validate | tail -2 ; echo "exit=$?"
Result: PASS (629 warnings)
Schemas: common@0.3.0 (on-disk), ...
exit=0
```

276 → 0. My earlier claim inferred "only `sysml2` *can* be on-disk" from the
observation "only `sysml2` *is* on-disk" — an inference reported as a measurement.

## …but the fix is still not to widen the vocabulary

Now that the option is known to exist, it should be declined on the merits rather
than on a false constraint. Appending `passing` / `not-implemented` / `planned` /
`partial` to a **vendored fork of the org-shared `common` schema** would:

1. **Silence a correct diagnostic.** rivet isn't being pedantic — it is reporting a
   genuine modeling error. The errors are the signal, not the noise.
2. **Fork spar's lifecycle vocabulary from every other PulseEngine repo**, silently,
   in a file nothing else reviews.
3. Be the textbook anti-pattern: *"fixing a red claim-check by softening the ledger
   instead of the claim."*

A green `rivet validate` bought that way is worth strictly less than the red one it
replaces.

## Proposed work

1. Migrate `artifacts/verification.yaml` entries from `type: feature` + verdict-in-
   `status` to `type: verification-verdict` (+ `verification-execution` where a run
   is being recorded), with `status` carrying a real lifecycle value.
2. Re-point the `verifies` / `verified-by` links at the new ids.
3. The success oracle is mechanical and already known: **`rivet validate` exit 0 with
   `common@0.3.0 (embedded)`** — i.e. green against the *unmodified* shared schema, no
   vendored file. That distinction is the whole point.
4. Only then is the `rivet-cli v0.4.3` pin in `ci.yml` liftable. (It is genuinely
   load-bearing today: CI's `Rivet validate (artifacts)` is green on v0.4.3 while
   0.28.0 reports 276 on the same tree — but the reason is this defect, not an
   unfixable one.)

## Scope note

~276 artifacts touched. This is a migration, not a patch, and it is the prerequisite
for #358 (Verification Gate scope stale) doing anything meaningful — a gate over
artifacts that are the wrong type can't select the right ones.

Related: #360, #368, #370 (release field on the wrong plane — same species: a field
in a place the tool doesn't read).

avrabe:
Closes #371.

Stacked on #368 (base is `fix/360-rivet-source-dedup`); GitHub will retarget to `main` when that merges. The two touch `artifacts/requirements.yaml` and so does #373 — whichever lands last needs a rebase.

## What this is

spar carried a private status vocabulary on **276 artifacts**. None of the four words is in the schema's allowed set, so every one was a `status-allowed-values` error and `rivet validate` could not run clean.

| from | to | n | lossy? |
|---|---|---|---|
| `passing` | `implemented` | 177 | no |
| `planned` | `proposed` | 40 | no |
| `not-implemented` | `proposed` | 51 | no |
| `partial` | `approved` | 8 | **yes** |

`passing` is the group worth naming. It is a verification **verdict** sitting in a lifecycle field, and a hand-edited YAML file cannot honestly assert that a test passes — only that the measure exists. That is what `implemented` says.

## The lossy eight

`partial` sits strictly between `approved` and `implemented`, and the vocabulary has no such rung. `approved` is the **conservative floor** — the strongest claim certainly true — rather than `implemented`, which would overstate.

Two of the eight (`RENDER-REQ-003`, `COVERAGE-GAPS`) already itemized their residue in prose and were left alone. The other five now say the residue is **not** itemized. That wording is deliberate and I want it reviewed as a choice, not skimmed as boilerplate: `lower.rs` has 54 `#[test]` functions, but STPA-REQ-036 requires *one test per lowering rule* — a mapping question a count cannot answer. Writing "54 tests exist" would be the necessary-but-not-sufficient trap. An explicit "not measured" is falsifiable; an invented figure is not, and inventing one is the exact failure #294 was reopened for.

## No new guard ships here — on purpose

rivet already enforces `status-allowed-values` at **ERROR** severity and `rivet validate` is already a required check, so the regression is mechanically blocked the moment this lands. A bespoke checker would add a weaker second oracle beside a working one.

This is the **opposite** of #370 / #373, which needed `tools/check_release_plane.py` precisely because rivet was *blind* to that defect.

## Verification — three planes that agree

1. **The migration proves its own edit.** Per file it asserts the artifact-id set is unchanged, every record is byte-identical outside `status`, and the text-rewrite count equals the semantic-move count. That last assertion is the load-bearing one: *"N lines rewritten"* is equally consistent with having rewritten N **prose** lines inside `description:` blocks while leaving N real statuses alone. Block scalars are tracked and skipped rather than `sed`'d through.
2. **The oracle flips.** `rivet validate` → exit 0, **0** error-severity diagnostics under `common@0.3.0 (embedded)`, down from 276. The diff is `276+/276-` on the status lines — a strict 1:1 swap, no structural change.
3. **The gate is live, not vacuous.** Re-introducing one off-vocabulary status reproduces **exactly one** `status-allowed-values` error; restoring it returns exit 0 with a byte-identical file. Distinct inputs → distinct verdicts. This is the property #327's withdrawn certificate did *not* have, so I no longer report an exit-0 without it.

## Reconciliation: why 303 raw hits but 276 errors

| | count | why |
|---|---|---|
| raw `status:` text matches, 9 committed files | 303 | |
| − `safety/requirements.yaml` | −21 | rivet **never loads it** — it is under no path in `rivet.yaml`'s `sources:` |
| − `safety/stpa/architecture.yaml` | −6 | nested in the `fields:` extension bag, where rivet cannot see them |
| **= rivet's error count** | **276** | ✅ exact |

Both remainders are **left in place and named** in `REQ-GUARD-STATUS-VOCAB-001` rather than silently absorbed into the total. The second one is the #370 defect one field over — `status:` instead of `release:` — and `tools/check_release_plane.py` from #373 would catch it by changing a single module-level constant (`FIELD = "release"`). That generalization is a follow-up, not this PR: doing it here would stack this branch three-deep and break the "vocabulary only" scope I posted to #371.

## Deliberately NOT in scope

Re-typing the 198 `TEST-*`/`VAL-*` measures to `unit-verification` and re-pointing `satisfies` → `verifies`. As measured on #371, those are ASPICE types requiring `sw-detail-design`/`sw-req`, of which spar has **zero** across 883 artifacts — adopting them would drag in SWE.1/SWE.2/SWE.3 and take feature-stage coverage from 194 requirements to 22. Separately scoped; the root friction is filed as pulseengine/rivet#748.

### DIFF
--- a/artifacts/architecture.yaml
+++ b/artifacts/architecture.yaml
@@ -456,7 +456,7 @@ artifacts:
 
   - id: ARCH-MCP-001
     type: design-decision
-    status: planned
+    status: proposed
     title: MCP server as CLI command with stdio JSON-RPC transport
     description: >
       The MCP server is a new spar CLI subcommand (spar mcp) that
@@ -503,7 +503,7 @@ artifacts:
 
   - id: ARCH-QUERY-001
     type: design-decision
-    status: planned
+    status: proposed
     title: "Path-expression query language: components.where().property() pattern"
     description: >
       Query expressions use a method-chain syntax over the instance model:
@@ -527,7 +527,7 @@ artifacts:
 
   - id: ARCH-KNOWLEDGE-001
     type: design-decision
-    status: planned
+    status: proposed
     title: Instance model is the knowledge base; query layer on top
     description: >
       The spar instance model (SystemInstance with its arena-allocated
@@ -555,7 +555,7 @@ artifacts:
 
   - id: ARCH-SOLVER-001
     type: design-decision
-    status: planned
+    status: proposed
     title: "NDS-layered hierarchical solving"
     description: >
       Deployment optimization uses 4 hierarchical layers analogous to NDS
@@ -659,7 +659,7 @@ artifacts:
 
   - id: ARCH-SOLVER-006
     type: design-decision
-    status: planned
+    status: proposed
     title: "Dual solver strategy (WASM heuristic + native exact)"
     description: >
       Two solver tiers. Tier 1 (WASM): pure Rust heuristics (FFD/BFD
--- a/artifacts/requirements.yaml
+++ b/artifacts/requirements.yaml
@@ -121,7 +121,11 @@ artifacts:
     description: >
       Trace connections through the hierarchy resolving across, up,
       and down patterns to produce end-to-end connection instances.
-    status: partial
+      Partially implemented: `partial` has no lifecycle equivalent, so
+      `approved` is the conservative floor, not an accurate position.
+      The unimplemented residue is deliberately not itemized here — see
+      REQ-GUARD-STATUS-VOCAB-001.
+    status: approved
     tags: [instance, as5506-ch14]
 
   - id: REQ-INST-004
@@ -226,7 +230,11 @@ artifacts:
       Enforce AADL naming rules, category restrictions, direction rules,
       connection rules, flow rules, binding rules, mode rules, and
       subcomponent rules. Target 50+ rules.
-    status: partial
+      Partially implemented: `partial` has no lifecycle equivalent, so
+      `approved` is the conservative floor, not an accurate position.
+      How many rules are actually enforced against the 50+ target is
+      deliberately not asserted here — see REQ-GUARD-STATUS-VOCAB-001.
+    status: approved
     tags: [analysis, legality]
 
   - id: REQ-ANALYSIS-008
@@ -309,7 +317,7 @@ artifacts:
     description: >
       Import cargo metadata to generate virtual AADL packages
       representing Rust crate architecture.
-    status: planned
+    status: proposed
     tags: [transform, rust]
 
   - id: REQ-TRANSFORM-003
@@ -318,7 +326,7 @@ artifacts:
     description: >
       Read WAC composition files and generate AADL system
       implementations with subcomponents and connections.
-    status: planned
+    status: proposed
     tags: [transform, wac]
 
   - id: REQ-TRANSFORM-004
@@ -327,7 +335,7 @@ artifacts:
     description: >
       Map wRPC transports (NATS, TCP, QUIC) to AADL bus components
       with deployment bindings.
-    status: planned
+    status: proposed
     tags: [transform, wrpc]
 
   - id: REQ-TRANSFORM-005
@@ -336,7 +344,7 @@ artifacts:
     description: >
       Map WASI P3 async functions, streams, and futures to AADL
       thread components with scheduling properties.
-    status: planned
+    status: proposed
     tags: [transform, wasi-p3]
 
   # ── STPA Analysis Coverage Records (STPA-REQ-022) ────────────────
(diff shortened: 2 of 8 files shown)


avrabe:
**Correction to the PR body: the base is now `main`, not `fix/360-rivet-source-dedup`.**

I opened this stacked on #368 for a clean diff. That was wrong, and wrong in a way worth recording because it is the exact defect class #372 is about.

`ci.yml` triggers on `pull_request: branches: [main]`. Based on another PR's branch, **no workflow triggered at all** — `total_count: 0`, zero runs. The PR presented as clean; `tot=0 pend=0 bad{none}` is indistinguishable from "all green" in every surface I check. I nearly reported this as awaiting-green when it was awaiting-nothing.

Retargeting to `main` **did not** fix it either: GitHub's default `pull_request` types are `opened, synchronize, reopened`, and changing a base is `edited`. It took a close/reopen to fire the event. Now at 7 check-runs across CI, Lean Proofs and Verification Gate.

Consequence for review: until #368 merges, this PR shows **3 commits** — #368's two plus this one — because that is genuinely the diff against `main`. Once #368 lands, GitHub recomputes and only `a3512eb` remains. The `rivet validate` exit-0 result holds either way; it was measured on top of #368's tree, which is what this branch contains.

Written up in #372 as an argument that the proposed standing check should also assert **every open PR has a non-zero required-context count** — no amount of auditing `ci.yml` catches a PR that never triggered it.

avrabe:

Verified on `origin/main` before closing, rather than assumed from the batch's commit list: Zero occurrences of `passing` / `not-implemented` / `in-progress` / `todo` remain in `artifacts/` or `safety/`; the lifecycle vocabulary is in use (317 `implemented`/`verified`, 4 `approved`).

PR #376, opened 2026-07-30
ci(release): gate the tag point — a tag behind main silently drops work (#366)

### LINKED ISSUE #366: Release tags can silently exclude finished work — happened in v0.31.0 and v0.33.0
## Summary

Tagging a release at a commit that is **behind `main`** silently drops finished
work from that release. Nothing in CI notices. I nearly did it cutting v0.35.0 —
and then found it has **already happened twice**, in v0.31.0 and v0.33.0.

## It is not hypothetical

| tag | tagged at | tagger time | landed on main before the tag but NOT in it | shipped instead in |
|---|---|---|---|---|
| **v0.31.0** | `98bc574` (the bump) | 2026-07-22 21:05:22 | `9557dd0` merged **21:00:06** — `fix(parser): reject C-style radix literals (0x/0b/0o) instead of silently parsing to 0` (#337/#343) | v0.32.0 |
| **v0.33.0** | `4cd5948` (the bump) | 2026-07-29 19:12:53 | `a2cee09` (16:08:00, #338/#349 target constants + linker map) and `9669870` (03:33:34, #348 rivet-validate determinism) | v0.34.0 |

v0.31.0 missed that parser fix by **five minutes and six seconds**. A fix whose
whole subject is "stop silently producing a wrong value" was itself silently
left out of a release.

The v0.35.0 near-miss would have been worse than a delay: tagging the bump
`9de1d8d` would have **shipped** `048d65b` (a codegen byte-layout certificate
later found to prove a tautology) and **excluded** `2af55b1`, the commit that
withdraws it.

## Why nothing catches it

The convention is to tag `vX.Y.Z` at its `chore(release): bump to vX.Y.Z`
commit. That is sound **only while the bump is still the tip**. Under the
batching policy — hold the tag, keep merging — the bump drifts backwards and
the convention starts pointing at a stale tree.

`check-versions` (`.github/workflows/release.yml:25`) compares the tag against
`Cargo.toml` and `vscode-spar/package.json`. Both read `0.35.0` at `9de1d8d`
*and* at `c12770a`, so the job passes identically on the right commit and the
wrong one. It is measuring the version window, not the tag point.

## My first proposed gate was wrong — recording it so it isn't re-proposed

I first proposed: enumerate `C..origin/main` and fail if any commit's
`Cargo.toml` still reads the tag's version. Invariant: *a commit carrying
version X.Y.Z that descends from the tag belongs to release X.Y.Z and was
excluded from it.*

**That invariant is false.** `Cargo.toml` reads `X.Y.Z` from the bump commit
until the *next* bump, and the tag sits somewhere inside that window. So the
version field spans two different populations:

```
  bump X.Y.Z ──────── tag vX.Y.Z ──────── bump X.Y.Z+1
       └── in release X.Y.Z ──┘└── in the NEXT release ──┘
       ←────── Cargo.toml reads X.Y.Z the whole way ──────→
```

The field cannot separate them, and the thing that separates them — which side
of the tag a commit is on — is exactly what the check was trying to determine.
Circular.

Its control case proved it: run against v0.34.0 at its own tag point and it
flags `048d65b` and `69e2b30`, which legitimately shipped in v0.35.0. It is
also **non-idempotent** — re-running it against v0.35.0 right now flags
`cd4d429` (#359), whose `Cargo.toml` reads `0.35.0` only because no bump has
happened since. A gate whose verdict changes as unrelated commits land is not
a gate.

I also hypothesised that filtering by tagger date would rescue it. It does not:
under design 1 the stragglers predate the tagger date anyway, because the tag
itself was held.

## The gate that actually works

Stop asking *"which release does this commit belong to"* and ask the question
that was really at stake: **"is anything that was already finished being left
out, and did you say so?"**

Given tag `vX.Y.Z` at commit `C`, with tagger date `T`:

1. `C` not an ancestor of `origin/main` → **skip** (release-branch tag).
2. Enumerate `C..origin/main`, **restricted to commits dated before `T`**.
   A commit merged after the tag was cut could not have been in it.
3. Empty → **pass**.
4. Non-empty and the tag message states a tag point → **pass**, listing them.
5. Non-empty and it does not → **fail**, listing them.

Step 2 is what makes it both race-proof (a merge landing seconds after the tag
push cannot fail the job) and **idempotent** (the verdict never changes on
re-run, because both inputs are immutable).

```bash
TAG="${GITHUB_REF#refs/tags/}"
C="$(git rev-parse "$TAG^{commit}")"
T="$(git for-each-ref "refs/tags/$TAG" --format='%(taggerdate:iso-strict)')"
git fetch --no-tags origin main

git merge-base --is-ancestor "$C" origin/main || { echo "Release-branch tag, skipping."; exit 0; }

EXCLUDED="$(git log --format='%h %s' --before="$T" "$C..origin/main")"
[ -z "$EXCLUDED" ] && { echo "PASS: nothing predating the tag was left out."; exit 0; }

if git tag -l "$TAG" --format='%(contents)' | grep -qiE 'tag[ -]point'; then
  echo "PASS (acknowledged):"; printf '%s\n' "$EXCLUDED" | sed 's/^/  excluded: /'; exit 0
fi
echo "::error::Commits were on main before this tag and are not in it:"
printf '%s\n' "$EXCLUDED" | sed 's/^/::error::  /'
echo "::error::Either tag the tip, or state the tag point in the tag message."
exit 1
```

Needs `fetch-depth: 0` on that job's `actions/checkout@v4`; the default of 1
has no history to walk.

## Verified against every release tag, not just the motivating case

Run over all 42 `v0.*` tags, with `main`'s tip reconstructed at each tagger
date:

- **40 pass.** No false positives.
- **2 fail: v0.31.0 and v0.33.0** — both confirmed real above, independently of
  the gate, by comparing merge timestamps to tagger timestamps.
- **v0.34.0 passes via rule 4**: its message already says *"Tag point is the
  bump commit ff470a2 … the tag was deliberately held after the bump merged,
  under the batching policy."*

That last point matters: the acknowledgement convention **already exists** in
how these tags are written. The gate does not invent process, it enforces a
habit that is currently optional and was skipped exactly when it mattered.

## What this does not do — stated so it is not mistaken for coverage

- **It cannot verify the acknowledgement is true.** It greps the tag message
  for a stated tag point. A message saying "tag point: whatever" passes. It is
  a speed bump against autopilot, not a proof — its value is forcing the
  tagger to look at the list, which is precisely what did not happen for
  v0.31.0 and v0.33.0.
- It cannot catch tagging a commit that is too *new*. Different failure.
- It cannot catch work merged after the tag — nothing at tag time can, and
  step 2 deliberately stops trying.
- It assumes tagger dates and committer dates come from comparably-set clocks.
  They do here (one maintainer, one machine); a badly skewed clock would
  weaken step 2 in the *permissive* direction.

## Scope

CI plumbing, no rivet artifact — matching the #353/#363/#364 precedent; this
repo carries no `REQ-CI-*` ids. Flagged so the omission reads as a choice.

Suggested for v0.36.0 alongside #358.

avrabe:
Closes #366.

## What broke, twice

| tag | what it excluded | margin |
|---|---|---|
| v0.31.0 | `9557dd0` — `fix(parser): reject C-style radix literals (0x/0b/0o) instead of silently parsing to 0` | **5m06s** |
| v0.33.0 | two commits | — |

Both with `check-versions` green.

## Why `check-versions` could never have caught it

Not an oversight — a category error. That job measures the version **window**: it asserts the tag matches `Cargo.toml`, and `Cargo.toml` reads `X.Y.Z` from the bump commit until the *next* bump. That window spans both the commits a tag included and the ones it silently left out. The invariant is circular with respect to the question being asked.

## The gate I designed first was also wrong

Recorded in #366 so it isn't re-proposed. Asking *"which release does this commit belong to"* makes the verdict a function of what has merged **since** — a gate whose answer drifts as unrelated commits land is not a gate. Filtering by tagger date doesn't rescue it either: under that design the stragglers predate the tagger date anyway, because the tag itself was held.

## What it actually asks

> Is anything that was already finished being left out, and did you say so?

Given tag `vX.Y.Z` at commit `C` with tag date `T`:

1. `C` not an ancestor of `origin/main` → **skip** (release-branch tag)
2. enumerate `C..origin/main`, restricted to `--before=T`
3. empty → **pass**
4. non-empty **and** the tag message states a tag point → **pass**, listing them
5. non-empty and it does not → **fail**, listing them

Step 2's `--before=T` is doing two jobs. **Race-proof:** a merge landing seconds after the tag push postdates `T` and cannot fail the job. **Idempotent:** both inputs are immutable, so a re-run months later returns the same verdict.

## Evidence

**Backtest — all 42 `v0.*` tags, running the shipped script, not a transcription of it:**

```
39 PASS   1 PASS-ACK (v0.34.0)   2 FAIL (v0.31.0, v0.33.0)
```

The two fails are exactly the two independently-confirmed bad tags. **No false positives across 40 good tags.** v0.34.0 passes via rule 4 because its message already says *"Tag point is the bump commit ff470a2 … deliberately held after the bump merged, under the batching policy"* — the acknowledgement convention already exists in how these tags are written; the gate enforces a habit that was optional and got skipped exactly when it mattered.

The real failure output names the cause directly:

```
::error::v0.31.0 was cut at 98bc574c (2026-07-22T21:05:22+02:00), but 1 commit(s) were
::error::already on main before that and are not in the release:
::error::  9557dd0 2026-07-22 21:00:06 +0200 fix(parser): reject C-style radix literals ... (#343)
```

**Mutation test — is rule 4 load-bearing, or does v0.34.0 pass for some other reason?** Two scratch tags on the *identical* commit `ff470a2`, yielding the *identical* 8-commit excluded set, differing only in whether the message contains "tag point":

| tag message | verdict |
|---|---|
| "…deliberately held after the bump merged, under the batching policy." | **FAIL** (exit 1) |
| "**Tag point** is the bump commit ff470a2; held under the batching policy." | **PASS** (acknowledged) |

Distinct inputs → distinct verdicts. (Both scratch tags deleted; `git tag -l 'zz-*'` → 0.)

## Two deliberate structural choices

**It lives in `tools/ci/check-tag-point.sh`, not inline in the YAML.** A backtest of shell transcribed into a workflow proves the transcription works. Putting it in a file means the 42-tag backtest and the mutation test exercise the artifact that actually ships — the same lesson as the withdrawn #327 certificate, where an oracle that couldn't be run against distinct inputs certified nothing.

**It gates `create-release`, not the builds.** A red verdict stops everything outward-facing (`publish-vsix` already `needs: create-release`) while the maintainer still gets full build feedback to act on.

## What this does not do

Stated in the script header rather than left to be discovered:

- **It cannot verify the acknowledgement is true.** It greps for a stated tag point; `"tag point: whatever"` passes. A speed bump against autopilot, not a proof. Its value is forcing the tagger to *look at the list* — precisely what did not happen twice.
- It cannot catch tagging a commit that is too **new**. Different failure.
- It cannot catch work merged **after** the tag. Nothing at tag time can, and step 2 deliberately stops trying.
- It assumes tag and committer dates come from comparably-set clocks. They do here. A skewed clock weakens step 2 in the **permissive** direction — it under-reports, never over-reports.

## Scope

CI plumbing, **no rivet artifact** — matching the #353/#363/#364 precedent; this repo carries no `REQ-CI-*` ids. Flagged so the omission reads as a choice, not a miss. It also keeps this PR clear of `artifacts/requirements.yaml`, which #368/#373/#374 are all currently touching.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,6 +46,34 @@ jobs:
           fi
           echo "All versions match: $TAG_VERSION"
 
+  # ── Tag-point check (issue #366) ──────────────────────────────────────
+  #
+  # `check-versions` above cannot catch a tag placed BEHIND main: it measures
+  # the version *window* (Cargo.toml reads X.Y.Z from the bump commit until the
+  # next bump), and that window spans both the commits a tag included and the
+  # ones it silently left out. v0.31.0 missed a parser fix by 5m06s and v0.33.0
+  # excluded two commits — both with check-versions green.
+  #
+  # This job gates the PUBLISH, not the builds: it is wired into
+  # `create-release`'s `needs`, so a red verdict stops anything outward-facing
+  # while the maintainer still gets full build feedback to act on. The logic —
+  # and the four things it deliberately does not claim to catch — live in the
+  # script.
+  tag-point:
+    name: Verify tag point
+    runs-on: ubuntu-latest
+    steps:
+      # fetch-depth: 0 is load-bearing: the gate walks C..origin/main, and the
+      # default depth of 1 has no history to walk. fetch-tags makes the tag
+      # OBJECT (not just the commit) present, which is what carries the tagger
+      # date and the message the gate reads.
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+      - name: Check the tag does not silently exclude finished work
+        run: tools/ci/check-tag-point.sh
+
   # ── Cross-platform binary builds ──────────────────────────────────────
   build-binaries:
     name: Build ${{ matrix.target }}
@@ -401,7 +429,7 @@ jobs:
   # the compliance-report archive, both folded into SHA256SUMS.txt.
   create-release:
     name: Create GitHub Release
-    needs: [build-binaries, build-compliance, build-test-evidence, build-vsix, build-sbom, build-wasm]
+    needs: [tag-point, build-binaries, build-compliance, build-test-evidence, build-vsix, build-sbom, build-wasm]
     runs-on: ubuntu-latest
     # Restated at job-level: actions/attest-build-provenance + cosign
     # keyless OIDC need both `id-token: write` and `attestations: write`.
--- a/tools/ci/check-tag-point.sh
+++ b/tools/ci/check-tag-point.sh
@@ -20,33 +20,52 @@
 #
 # Given tag vX.Y.Z at commit C with tag date T:
 #
+#   0. tag is lightweight, not annotated           -> fail (see below)
 #   1. C not an ancestor of origin/main            -> skip (release-branch tag)
-#   2. enumerate C..origin/main, dated before T    -> the excluded set
+#   2. enumerate C..origin/main on the FIRST-PARENT
+#      spine, dated before T                       -> the excluded set
 #   3. empty                                       -> pass
-#   4. non-empty, tag message states a tag point   -> pass, listing them
+#   4. non-empty, tag message carries a
+#      `Tag-Point:` trailer                        -> pass, listing them
 #   5. non-empty, it does not                      -> fail, listing them
 #
-# Step 2's `--before=T` is what makes the gate both race-proof and idempotent.
-# Race-proof: a merge landing seconds after the tag push cannot fail the job,
-# because it postdates T. Idempotent: both inputs (the tag object, and the set
-# of commits dated before T that are reachable from main) are immutable, so a
-# re-run months later returns the same verdict. A gate whose answer changes as
-# unrelated commits land is not a gate.
+# WHY --first-parent, AND WHY THE OBVIOUS VERSION IS WRONG. A clean-room audit
+# refuted the first design, which enumerated every commit in `C..origin/main`
+# dated before T. `--before` filters on COMMITTER DATE — when a commit was
+# written — but the question is when it LANDED ON MAIN. Those differ, and the
+# gap is unbounded: a branch whose commits are dated the 3rd, merged on the 10th,
+# injects commits "dated before T" into a range they were not reachable from at
+# tag time. Consequences, both demonstrated against a scratch repo:
+#
+#   * NOT IDEMPOTENT. Same tag object, same tagged commit, byte-identical
+#     ->  PASS at tag time, FAIL when re-run after that branch merged. A gate
+#     whose verdict drifts as unrelated commits land is exactly what #366
+#     rejected in its first design; this reintroduced it one level down.
+#   * OVER-REPORTS, contradicting the soundness direction this header used to
+#     claim ("weakens in the PERMISSIVE direction — it under-reports, never
+#     over-reports"). It is a false accusation against a tagger who did tag the
+#     tip of main. That claim was wrong and is deleted.
+#
+# The first-parent spine fixes both. It contains exactly the squash and merge
+# commits, whose committer dates ARE their landing times, so `--before` becomes
+# exact rather than approximate. A late-merged branch contributes its merge
+# commit (dated at merge time, correctly after T) and not its old constituents.
+# Detection is unaffected: the v0.31.0 miss (9557dd0) is a squash commit on the
+# spine. This repo squash-merges today, which made the defect latent rather than
+# active — main carries 91 merge commits, so "latent" is not "impossible".
 #
 # WHAT THIS DOES NOT DO — stated so it is not mistaken for coverage:
 #
-#   * It cannot verify the acknowledgement is TRUE. It greps the tag message for
-#     a stated tag point; a message saying "tag point: whatever" passes. This is
-#     a speed bump against autopilot, not a proof. Its value is forcing the
-#     tagger to look at the list — precisely what did not happen for v0.31.0 and
-#     v0.33.0.
+#   * It cannot verify the acknowledgement is TRUE. It requires a `Tag-Point:`
+#     trailer; a trailer saying "Tag-Point: whatever" passes. This is a speed
+#     bump against autopilot, not a proof. Its value is forcing the tagger to
+#     look at the list — precisely what did not happen for v0.31.0 and v0.33.0.
 #   * It cannot catch tagging a commit that is too NEW. Different failure.
 #   * It cannot catch work merged AFTER the tag. Nothing at tag time can, and
 #     step 2 deliberately stops trying.
-#   * It assumes tag dates and committer dates come from comparably-set clocks.
-#     They do here (one maintainer, one machine). A badly skewed clock would
-#     weaken step 2 in the PERMISSIVE direction — it under-reports, never
-#     over-reports.
+#   * On the first-parent spine it does not see commits that reached main other
+#     than through a merge or squash — a direct push to main. Branch protection
+#     forbids those.
 #
 # Usage:
 #   tools/ci/check-tag-point.sh v0.36.0     # explicit (backtesting, local runs)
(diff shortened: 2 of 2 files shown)


avrabe:
## Clean-room audit: this gate could be switched off silently, three ways

Same cold subagent, same instruction — refute, don't confirm. It confirmed A1
(the 42-tag backtest), A2 (v0.31.0 excluded `9557dd0`, merged **5m16s** before
the tag — the header said 5m06s) and A3. It refuted the rest, and one finding is
severe enough that merging this as written would have shipped a gate that
disarms on the sloppier path.

### 1. A lightweight tag made the gate vacuous — unconditional PASS

`creatordate` on a lightweight tag is the **tagged commit's own committer
date**, so `--before=$T` asks for commits descended from `C` that predate `C` —
on a linear history, always empty. Demonstrated on this repo, same commit
`945799a` tagged both ways:

```
annotated   -> ::error:: ... 13 commit(s) were already on main   rc=1
lightweight -> PASS: nothing predating the tag was left out      rc=0
```

Compounding: `%(contents)` on a lightweight tag returns the **commit** message,
so the acknowledgement could be satisfied by text the tagger never wrote.

The trigger is `git tag v0.36.0` instead of `git tag -s v0.36.0`. **The gate
switched itself off on exactly the mistake that most needs gating.** My header
called this "a slightly weaker check" — it was no check. Now rejected outright.

### 2. The acknowledgement disarmed itself

Rule 4 grepped the whole tag message for `tag[ -]point`. The commit that *adds
this gate* has the subject `ci(release): gate the tag point — …`, and release
notes here list commit subjects verbatim (v0.35.0's do). **The first release
whose notes quoted that subject would auto-acknowledge.** Reproduced: a
realistic CONTENTS block passed with zero acknowledgement.

Now anchored to the start of a line — a quoted subject sits mid-line after a
short hash; an acknowledgement is something the tagger *began a line with*. I
tried trailer-only (`Tag-Point:`) first and rejected it: it fails v0.34.0, whose
"Tag point is the bump commit ff470a2" is specific and correct. A gate that
rejects real acknowledgements teaches people to route around it.

### 3. Not idempotent — and it over-reported, contradicting my soundness claim

`--before` filters **committer date** (when a commit was *written*); the question
is when it **landed on main**. A branch dated the 3rd, merged the 10th, injects
commits "dated before T" into a range they weren't reachable from at tag time.
In a scratch repo — byte-identical tag object, only `main` moved:

```
RUN 1 (at tag time)      PASS   rc=0
RUN 2 (after the merge)  ::error:: 1 commit(s) were ...   rc=1
```

That is the verdict-drift #366 explicitly rejected in its first design,
reintroduced one level down. And it's an over-report — a **false accusation**
against a tagger who *did* tag the tip of main — refuting the header's claim to
err "in the PERMISSIVE direction". That sentence is deleted rather than softened.

**Fix: `--first-parent`.** The spine holds exactly the squash and merge commits,
whose committer dates *are* their landing times, so `--before` becomes exact
instead of approximate. A late-merged branch contributes its merge commit (dated
after T, correctly excluded), not its old constituents. Same scratch repo now
gives **PASS / PASS**.

Detection is untouched: v0.31.0's missed commit `9557dd0` is a squash commit on
the spine. This repo squash-merges, which made the defect latent — but main
carries **91** merge commits, so latent is not impossible.

### Verification

Backtest over all 42 tags, unchanged end-to-end: **39 PASS, 1 acknowledged
(v0.34.0), 2 FAIL (v0.31.0, v0.33.0)** — the two real misses #366 was filed for.
All three counterexamples above now exit 1 (measured directly, not through a
pipe). Scratch tags cleaned up; `git tag -l 'zz-fix-*'` → 0.

The holes closed without moving a single verdict on real history.

avrabe:

Verified on `origin/main` before closing, rather than assumed from the batch's commit list: `.github/workflows/release.yml` carries the tag-point guard (6 matching lines).

Issue #390, opened 2026-08-05
taprio GCL readback ignores sched-entry cmd (S/H/R treated alike)


avrabe:
Split out of the taprio readback fix (489c834) so the decision is recorded rather than silently deferred.

`extract_taprio_gcl` in `crates/spar-trace-topology/src/fixtures/transform.rs` reads every entry of `options.schedule` as a gate-control-list entry, using only `gatemask` and `interval`. It never looks at `cmd`.

iproute2 emits `cmd` as one of three values (`tc/q_taprio.c`, `entry_cmd_to_str`):

- `S` — SetGates, the entry carries a gate mask
- `H` — Hold, for frame preemption
- `R` — Release, likewise

Only `S` entries are gate entries in the 802.1Qbv sense. `H`/`R` belong to 802.1Qbu/802.3br preemption and their gate mask is not a schedule state the Qcc YANG `gate-control-list` should carry.

**Why it has not bitten.** `bin/gen-fixtures.rs` installs only `sched-entry S` entries, so every entry read back is an `S`. The distinction is unobservable on the current fixture topology, which is exactly why it should be written down instead of trusted to stay that way — a future fixture that exercises preemption would silently get `H`/`R` masks folded into its GCL.

**Not fixed in 489c834 deliberately**: that commit corrects the JSON key names, and changing which entries are accepted is a semantic change to the emitted Qcc YANG. Bundling them would have made the mutation-test evidence for the shape fix ambiguous.

**Proposed:** filter to `cmd == "S"`, and decide explicitly what a non-`S` entry means — skipped, or an error naming the unsupported command. An entry with no `cmd` at all should probably be an error rather than assumed to be `S`, on the same principle as the rest of that function: absence should not read as the expected case.

Blocked on nothing; wants a fixture with a preemption entry to be a real test rather than an assertion about code that cannot run.

Refs #362, #365

avrabe:
**spar issue-hunt run — picked this item, PR opened.**

Overlap guard: zero open PRs at start, main CI green (`e6e6355`). Picked #390 as a clean-scoped bug with a concrete oracle.

**What shipped (into PR, CI pending):** [#394](https://github.com/pulseengine/spar/pull/394), branch `feat/taprio-cmd-filter-390`.

`extract_taprio_gcl` now dispatches on the sched-entry `cmd`:
- `S` → gate-control entry (unchanged extraction)
- `H` / `R` → skipped (valid preemption entry, not folded into the GCL)
[...]

Issue #383, opened 2026-07-30
Format check is green on unformatted tracked code: cargo fmt --all covers 1 of the repo's 4 workspaces


avrabe:
The required `Format` check runs `cargo fmt --all -- --check`. `--all` means "all members of *this* workspace" — not "all crates in the repo". The repo has **four** workspaces:

```
$ grep -rn "^\[workspace\]" --include=Cargo.toml .
Cargo.toml:1
codegen-exec-oracle/Cargo.toml:13
codegen-kiln-oracle/Cargo.toml:20
fuzz/Cargo.toml:13
```

The three nested ones each carry their own `[workspace]` table, which makes them separate workspaces. `cargo fmt --all` from the root never opens them.

## This is live on HEAD, not hypothetical

```
$ cargo fmt --all -- --check                              ; echo $?
0
$ cargo fmt --manifest-path fuzz/Cargo.toml -- --check    ; echo $?
1                                        # 13 diff lines
$ cargo fmt --manifest-path codegen-exec-oracle/Cargo.toml -- --check ; echo $?
0
$ cargo fmt --manifest-path codegen-kiln-oracle/Cargo.toml -- --check ; echo $?
0
```

The drift is in a tracked file, `fuzz/fuzz_targets/fuzz_codegen_roundtrip.rs:32`:

```diff
-    let sf = spar_base_db::SourceFile::new(
-        &db,
-        "fuzz.aadl".to_string(),
-        SEED_AADL.to_string(),
-    );
+    let sf = spar_base_db::SourceFile::new(&db, "fuzz.aadl".to_string(), SEED_AADL.to_string());
```

So a required status check named `Format` has been reporting green over unformatted tracked Rust. Nothing in CI has ever formatted `fuzz/`.

## Why it matters more than the diff

`Format` is one of only five required contexts that survive `code=false` in the change filter, so it is one of the few things always running — and it has a hole in it. The scope mismatch is silent by construction: `--all` *sounds* exhaustive, and cargo reports success for the subset it did check. Same shape as #381: the gate answered a narrower question than its name implies, and the narrow answer looks identical to the broad one.

## Fix

Extend the `fmt` job to cover every workspace in the repo, and — more importantly — make the *set* self-maintaining so a fifth workspace can't be added silently:

```yaml
- name: Format (all workspaces)
  run: |
    set -euo pipefail
    mapfile -t manifests < <(grep -rl '^\[workspace\]' --include=Cargo.toml . | sort)
[...]

Issue #381, opened 2026-07-30
Mutation Testing gate reads a path cargo-mutants never writes — 4 months of vacuous green (210 survivors reported as 0)


avrabe:
`Mutation Testing` is a required status check. It runs for ~3 hours, and it has
never once read its own results. It reports `Surviving mutants: 0` on every run,
including runs where 210 mutants survived a threshold of 142.

## The bug

```
cargo mutants -p spar-analysis --timeout 120 --jobs 4 --output mutants-out -- --lib || true
```

`--output DIR` tells cargo-mutants **which directory to create `mutants.out` in**.
The report therefore lands at `mutants-out/mutants.out/`. The gate reads:

```
MISSED=$(grep -c '^MISSED' mutants-out/caught.txt 2>/dev/null || echo 0)
if [ -f mutants-out/missed.txt ]; then
  MISSED=$(wc -l < mutants-out/missed.txt | tr -d ' ')
fi
```

Neither path has ever existed. `grep` fails, stderr is discarded, `|| echo 0`
supplies `MISSED=0`; the `-f` test is false so the real count is never read;
`0 -gt 142` is false; exit 0.

## Evidence

Artifact `mutants-report` (id 8773523705) from run 30563879100 — the one
`Mutation Testing` run in the last 25 that reached `success` — on `main` @
`cd4d429`. It is uploaded with `path: mutants-out/`, so **the artifact root is
the contents of `mutants-out/`**. That root contains exactly one entry:

```
mutants.out/
├── caught.txt      603 lines
├── missed.txt      210 lines
├── unviable.txt    793 lines
├── timeout.txt       2 lines
└── outcomes.json   total_mutants 1608
```

`outcomes.json` agrees line-for-line: `missed 210, caught 603, unviable 793,
timeout 2`. **The job's own uploaded artifact is the proof that the job cannot
read it** — the upload takes the whole tree, the gate looks one level too
shallow.

The gate printed `Surviving mutants: 0` / `within threshold (142)` for that same
run. I reproduced the mechanism by executing the step body locally: with the
directory absent it prints `Surviving mutants: 0` and exits 0; with a real
200-line `missed.txt` at the path it expects, it exits 1. The logic is fine. The
path is wrong.

## How long

`git log -L` shows the read path byte-identical since 563eb8c (2026-03-22
06:43:51 +0100) — the commit titled *"mutation testing as hard gate — zero
surviving mutants required"*. Four months. It was never right; there is no
regression to bisect.

[...]

PR #395, opened 2026-08-06
feat(guard): clippy every workspace, and assert the fuzz target set (#383, #390)

avrabe:
**REQ-GUARD-GATE-EVIDENCE-002 (g) and (h)** — obligations 2 and 3 of v0.36.0's bug-fix track (plan #392, first obligation #393).

Both are the same shape: a whole-repo tool whose scope is smaller than its name claims, with the shortfall rendering as the ideal reading.

## (g) Clippy linted 1 of 4 workspaces

`cargo clippy --workspace --all-targets` means *"all members of **this** workspace"* — the same scope `cargo fmt --all` had.

```
Cargo.toml   codegen-exec-oracle/   codegen-kiln-oracle/   fuzz/
```

Three workspaces have been unlinted behind a **required** check reporting green. The fmt half of #383 was fixed and this half left — which is exactly why the requirement says *"every whole-repo tool uses discovery, not just the one that was caught"*.

`check_clippy_workspaces.py` **reuses** `check_fmt_workspaces.discover` and its `check()`; it does not copy them. Tool wording is injected (`tool`/`verb`/`fix_hint`) with fmt defaults, so the fmt gate is behaviourally byte-identical — verified: its 10-case self-test still passes and it still prints `4 workspace(s) checked, 4 formatted`. Copying would put discovery in two files, and two copies that drift apart is *how #381 happened*.

**Measured before wiring** — all four workspaces are already clippy-clean under `-D warnings` (run individually, exit 0 each), so this is not a latent red. Live: `4 workspace(s) checked, 4 lint-clean.`

The discriminating self-test case is the runner's exit code: the **same** tree passes when the runner reports clean and **fails** when it reports dirty. A wrapper that ignored its runner — how this gate would go vacuous — passes the first and fails the second.

## (h) The fuzz target list was never compared to its declaration

`fuzz/Cargo.toml` declares `[[bin]]` harnesses; `fuzz-nightly.yml` runs a hand-written `matrix.include`. The `include` form is deliberate and stays (only `fuzz_scheduler_solver` needs `-max_len=128`; a flat list would impose that cap on two healthy fuzzers, #361). Its cost is two lists nothing compares.

Drop either side and the nightly stays green **having fuzzed a strict subset** — no error, no warning, no smaller number in the log, because success is per-leg and nobody counts the legs. Nightly is where this hides longest: `fuzz_scheduler_solver` was red **97/97 runs** from 2026-04-25 before anyone looked, and a silently *absent* leg is harder to notice than a loudly failing one.

Set equality **both directions** — declared-but-not-run is a harness never exercised; run-but-not-declared is a leg that cannot build and fails looking like a toolchain problem. Either side **empty exits 2**: a parse that read nothing is a broken scan, not perfect agreement about nothing — the one way this check could itself go vacuous. 7-case self-test, two file reads, no build.

## Two things the gate surfaced by being run

Included rather than reverted:

- **`codegen-*/Cargo.lock` pinned `spar-analysis 0.25.0`** against a workspace at **0.35.0** — ten minor versions stale, because nothing routinely built there to refresh them. Otherwise every contributor who runs the new gate gets the same dirty diff.
- **`codegen-*/target/` and `fuzz/target/` were not gitignored** — `.gitignore:1` is `/target`, root-anchored. The gap never showed because nothing built there. It does now.

## Verified

- **9/9** guardrail self-tests pass on this tree
- both new gates run green for real
- release-plane and human-scoped gates still pass on the edited artifacts
- workflow + artifact YAML parse

Refs #383, #390

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,16 @@ jobs:
         with:
           components: clippy
       - uses: Swatinem/rust-cache@v2
-      - run: cargo clippy --workspace --all-targets -- -D warnings
+      # REQ-GUARD-GATE-EVIDENCE-002 (g), the clippy half of #383.
+      # `cargo clippy --workspace` means "all members of THIS workspace" — the
+      # same scope `cargo fmt --all` had — so this linted 1 of the repo's 4
+      # workspaces and reported green for all of them. The checker reuses
+      # check_fmt_workspaces' discovery rather than copying it: one definition,
+      # two callers, because a fact duplicated in two files is what made #381
+      # possible. Self-test first, same as every guardrail here.
+      - name: Clippy-scope guardrail self-test
+        run: tools/check_clippy_workspaces.py --self-test
+      - run: tools/check_clippy_workspaces.py
 
   # ── Tests ─────────────────────────────────────────────────────────────
   test:
@@ -538,6 +547,16 @@ jobs:
       # every filter is vacuous.
       - name: Verification-filter guardrail
         run: tools/check_verification_filters.py
+      # REQ-GUARD-GATE-EVIDENCE-002 (h). fuzz-nightly names its targets in a
+      # hand-written matrix `include` (each carries its own extra_args, #361),
+      # so the declared set and the run set are maintained separately and
+      # nothing compared them. A target dropped from either side leaves the
+      # nightly green having fuzzed a strict subset — no error, no warning, no
+      # smaller number in the log. Cheap: two file reads, no build.
+      - name: Fuzz-target guardrail self-test
+        run: tools/check_fuzz_targets.py --self-test
+      - name: Fuzz-target guardrail
+        run: tools/check_fuzz_targets.py
       - name: Release-plane guardrail
         run: |
           tools/check_release_plane.py \
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,12 @@ __pycache__/
 # far enough to produce either one.
 /tools/fixture-vm/nixos.qcow2
 /tools/fixture-vm/result
+
+# Nested-workspace build output. `/target` above is ROOT-ANCHORED, so it does
+# not cover the three sibling workspaces; before the clippy-scope guardrail
+# (REQ-GUARD-GATE-EVIDENCE-002 g) nothing routinely built in them, so the gap
+# never showed. It does now — running clippy repo-wide leaves untracked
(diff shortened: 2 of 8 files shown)


avrabe:
## Red board here is a GitHub Actions outage, not this PR

Recording the diagnosis so nobody re-reads these as code failures. **Not one of the non-green checks has executed a single step of real work.**

GitHub's own status page, checked independently:

```
overall: Partial System Outage
  Actions          major_outage
INCIDENT: Incident with Actions [investigating]  2026-08-06T15:22
```

Every failure on this PR falls inside that window (first attempt 15:10).

### The three non-green checks, and why none is a verdict

| check | conclusion | evidence |
|---|---|---|
| `Lean proof typecheck` | **failure** | died in `Set up job`: `Failed to resolve action download info. Error: Service Unavailable` — before checkout |
| `Code Coverage` | **cancelled** | no log exists at all (`log not found`) — never ran a step |
| `Security Audit (RustSec)` | **cancelled** | same |

`cancelled` is not `failure`, and both matter differently. The cancellations are **self-inflicted and my fault**: this workflow sets

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

so back-to-back `gh run rerun --failed` across the two runs killed in-flight jobs from the same group. Re-runs were staggered afterwards; the outage outlasted them.

### One hypothesis I had, checked and wrong

I suspected the `Security Audit` red was mine — this PR refreshes `codegen-*/Cargo.lock`, which plausibly pulls a version carrying an advisory. It is not: the job never reached `cargo audit`, or any step. Worth stating because acting on that hypothesis would have meant reverting a legitimate fix (those lockfiles pinned `spar-analysis 0.25.0` against a workspace at `0.35.0`) to chase a failure that never touched the code.

### What IS verified, locally

- **9/9** guardrail self-tests pass on this tree
- both new gates run green for real — `4 workspace(s) checked, 4 lint-clean` and `All 3 declared fuzz targets are run`
- all four workspaces measured clippy-clean under `-D warnings` **before** wiring, so (g) is not a latent red
- `check_fmt_workspaces` behaviourally unchanged by the parameterisation: 10/10 self-test, still `4 workspace(s) checked, 4 formatted`

### Next

Waiting for the incident to clear, then one clean re-run. Not merging on an outage-red board, and not re-running again while Actions is in `major_outage` — that just burns runner time on the self-hosted pool.

